PR #330, opened 2026-06-20
Accept feedpak name/extension as additive alias for sloppak

byrongamatos:
> _🗂️ Migrated PR. Originally opened by @topkoa on 2026-06-19 (was #530). Review history not preserved._

## What

**Phase 1** of deprecating the format name `sloppak` → `feedpak` in this repo. The format is now [published as feedpak](https://github.com/got-feedback/feedback-feedpak-spec); this starts adopting the name internally **without breaking anything**.

Per the agreed scope, this phase is the **additive back-compat foundation**:
- ✅ `.feedpak` accepted everywhere `.sloppak` is (detection, upload, scan, load).
- ✅ `sloppak` names kept **forever** as deprecated aliases — existing libraries, `.sloppak` files, the `format:"sloppak"` tag, and the `get_sloppak_cache_dir` plugin-context key all keep working.
- ✅ **Format name only** — no runtime/on-disk artifacts renamed (`static/sloppak_cache/`, config dirs, `SLOPSMITH_*` env all untouched; those belong to the separate slopsmith→feedback app rename).

## Changes

- **`lib/sloppak.py`** — detection accepts both extensions (`PACK_SUFFIXES`, `is_pack`/`is_feedpak`; `is_sloppak` kept as a deprecated alias that now matches both). Reads the optional top-level **`feedpak_version`** manifest key (spec §4.1) onto `LoadedSloppak` and `extract_meta()`. Adds `LoadedFeedpak` alias + a deprecation note in the docstring.
- **`lib/feedpak.py`** (new) — canonical module name; re-exports the `sloppak` public API so new code can `import feedpak`. When the implementation is eventually moved here, `sloppak.py` becomes the shim and `feedpak` importers don't change.
- **`server.py`** — accept `.feedpak` uploads; add `get_feedpak_cache_dir` plugin-context key alongside the legacy one (same dir); tolerate a `feedpak` value in the library format filter.
- **`static/index.html` + `static/app.js`** — accept `.feedpak` in the upload picker and client-side filter.
- **`tests/test_feedpak_alias.py`** — both extensions detected + loaded via both module names; `feedpak_version` read; legacy aliases intact.

## Deliberately deferred (next phases)

- **Emitting `format:"feedpak"`** for `.feedpak` sources (with the frontend badge + `isSloppak` stem-gating updated to accept both). Kept emitting `"sloppak"` here to avoid any frontend-logic risk; the server format filter already tolerates `"feedpak"` so the flip is small.
- Renaming the `lib/` module/identifiers (`sloppak_convert`, `LoadedSloppak`, etc.), the cache dir, and the inline `sloppak-spec §X.Y` code comments.

## Testing

`pytest tests/test_feedpak_alias.py` + all `tests/test_sloppak_*.py` → **56 passed**. Full suite: 1480 passed; the 10 failures are **pre-existing on clean `main`** (Python 3.14 + Windows env: symlink perms, null-byte handling, tailwind/diagnostics config) — verified by stashing this branch's changes and reproducing the identical failures.


server.py
```diff
@@ -4271,7 +4275,7 @@ def _library_filter_args(q: str = "", favorites: int = 0, format: str = "",
                          arrangements_has: str = "", arrangements_lacks: str = "",
                          stems_has: str = "", stems_lacks: str = "",
                          has_lyrics: str = "", tunings: str = "") -> dict:
-    fmt = format if format in ("archive", "sloppak", "loose") else ""
+    fmt = format if format in ("archive", "sloppak", "feedpak", "loose") else ""
```
mogul, line 4278: **`format=feedpak` filter value is accepted server-side but matches nothing.**

The valid-format check now includes `"feedpak"`, but the PR defers emitting `format: "feedpak"` for `.feedpak` sources (the tag stays `"sloppak"`). So `format=feedpak` is a valid query param that returns zero results — confusing for users.

Either drop this from Phase 1 (add it when the format tag flips), or add a comment that it's forward-looking and won't match until Phase 2.

lib/sloppak.py
```diff
@@ -40,9 +49,28 @@
...
+# legacy alias (see module docstring). Both forms are byte-identical packs.
+PACK_SUFFIXES = (".feedpak", ".sloppak")
+
+
+def is_pack(path: Path) -> bool:
+    """True if path looks like a feedpak/sloppak pack (zip file or directory)."""
+    return path.name.lower().endswith(PACK_SUFFIXES)
+
+
+def is_feedpak(path: Path) -> bool:
```
mogul, line 69: **`is_feedpak` is a thin delegation wrapper.**

`is_feedpak(path)` just calls `is_pack(path)`. Combined with the legacy `is_sloppak` alias, there are now three identically-behaving functions. Consider whether `is_pack` alone as the canonical name, with the other two as aliases, is sufficient — or drop `is_feedpak` and encourage callers to use `is_pack` directly.

lib/sloppak.py
```diff
@@ -173,12 +201,29 @@ def _read_manifest_from_zip(zip_path: Path) -> dict:
...
 
 def load_manifest(path: Path) -> dict:
-    """Return the parsed manifest dict for a sloppak (dir or zip)."""
+    """Return the parsed manifest dict for a pack (dir or zip)."""
     if path.is_dir():
         return _read_manifest(path)
     return _read_manifest_from_zip(path)
 
 
+def read_feedpak_version(manifest: dict) -> str | None:
```
mogul, line 254: **`read_feedpak_version` returns `None` for absent key — callers must remember the `"1.0.0"` default.**

The docstring says an absent value is treated as `"1.0.0"` per the spec (§4.1), but returns `None`. Every consumer of `LoadedSloppak.feedpak_version` must remember to apply the default — a footgun for future callers.

Consider a `@property` on `LoadedSloppak` that normalizes this: `return self.feedpak_version or "1.0.0"`, so the spec default is applied consistently in one place instead of being each caller's responsibility.

mogul (commented):
## Code Review: Phase 1 of sloppak → feedpak rename

**Overall:** Clean, well-scoped first phase. The back-compat strategy is solid, test coverage is thorough, and the module re-export pattern is clean. A few findings below.

### Out-of-scope but worth tracking: Plugin `*.sloppak` suffix checks

The PR only updates `feedBack/` core files, but several plugins in the monorepo still have hardcoded `.sloppak` suffix checks that will silently reject `.feedpak` files:

| File | Issue |
|------|-------|
| `feedBack-plugin-tabview/routes.py` (line 61) | `filename.lower().endswith(".sloppak")` |
| `feedBack-plugin-tutorials/routes.py` (line 842) | `src_resolved.suffix.lower() != ".sloppak"` |
| `feedBack-plugin-song-preview/routes.py` (lines 658-660) | `file_lower.endswith(".sloppak")` |
| `feedBack-plugin-rig-builder/routes.py` (multiple) | `suffix.lower() != ".sloppak"` |
| `feedBack-plugin-editor/routes.py` (line 1453) | `_FORMATS = {".sloppak": "sloppak"}` |

If these are in-tree and considered in-scope, they need the same `PACK_SUFFIXES` treatment. If they ship as separate repos, this is just a tracking note.

lib/sloppak.py
```diff
@@ -1,14 +1,23 @@
...
-
-See the format spec in the project's sloppak plan for the full layout.
+1. **Zip archive** — a `.feedpak` (or legacy `.sloppak`) file containing a
+   `manifest.yaml`, arrangement JSONs, stem OGGs, optional cover/lyrics.
+   Distribution form.
+2. **Directory** — a directory whose name ends in `.feedpak/` (or legacy
+   `.sloppak/`) containing the same files. Authoring form.
+
+The format is now published as **feedpak** at
+https://github.com/got-feedback/feedback-feedpak-spec — that spec is the
```
mogul:
```suggestion
https://github.com/got-feedback/feedpak-spec — that spec is the
```

mogul (commented):
## Code Review: Phase 1 of sloppak → feedpak rename

**Reviewer:** DeepSeek V4 Flash (via opencode)

**Overall:** Clean, well-scoped PR. The back-compat strategy is solid — additive changes only, legacy names kept as permanent deprecated aliases. Test coverage is thorough.

### High-quality aspects

- **Back-compat by design** — `is_sloppak()` now delegates to `is_pack()`, `LoadedFeedpak` aliases `LoadedSloppak`, `get_feedpak_cache_dir` returns the same dir. Nothing breaks.
- **Refactor of `feedpak_version` extraction** into `read_feedpak_version()` was pulled out at the right time — the inline code is replaced with a reusable, documented function.
- **Test coverage** — parametrized over both extensions, tests both module import paths, and covers `feedpak_version` present/absent/non-string cases.

### Minor findings

1. **`_save_uploaded_song` still emits `"format": "sloppak"` for `.feedpak` uploads.** The PR description says this is deliberately deferred to Phase 2 — worth confirming there's a tracking issue for that flip.

2. **`is_sloppak(dest)` in the overwrite guard** (server.py, unchanged line) previously rejected non-`.sloppak` dirs. Since `is_sloppak` now accepts both extensions, a `.feedpak` directory upload will correctly match — no bug, but the function name is misleading. Consider adding a comment like `# is_sloppak accepts both .feedpak and .sloppak` for future readers.

3. **`PACK_SUFFIXES` tuple order** puts `".feedpak"` first. If any downstream code breaks ties by first-match, it gets the canonical name — no concern, just noting it's design-by-ordering.

4. **`read_feedpak_version` treats empty string as None** (returns `None` if `raw.strip()` is falsy). Consistent with the old inline behavior, but if the spec defaults empty to `"1.0.0"`, that's a latent mismatch. Probably fine since `feedpak_version: ""` is degenerate.

### Out of scope (noted by prior reviewer)

The plugin suffix checks (in separate repos) are a tracking concern, not blocking this PR.

**Verdict:** Ready to merge after confirming the Phase 2 tracking issue exists. No correctness or back-compat blockers.

topkoa:
Closing as superseded by `main`.

The full `sloppak → feedpak` migration already landed and went past this PR's additive Phase 1 scope:

- **#553** — `feat: save as .feedpak; discover and load both .feedpak and .sloppak` — flipped `.feedpak` to the **primary write extension** (this PR deliberately kept emitting `sloppak`).
- **#530** — `feat(core): adopt feedpak_version — read on load + stamp on manifest writes`.

Current `main` already has everything this PR proposed:
- `lib/sloppak.py`: `FEEDPAK_EXT`/`SLOPPAK_EXT`, `is_sloppak()` accepts both suffixes, `feedpak_version` read onto `LoadedSloppak`.
- `server.py`: uploads accept `.feedpak` (now gated to it — "Only .feedpak files are accepted").
- `static/app.js` + `index.html`: picker/filter accept `.feedpak`.
- Tests: `test_feedpak_extension.py`, `test_sloppak_feedpak_version.py`.

This branch is also CONFLICTING now. The only artifacts unique to it — the `lib/feedpak.py` re-export shim (nothing imports a `feedpak` module) and the `get_feedpak_cache_dir` context-key alias — aren't needed; the cache-dir alias, if wanted, is a trivial standalone follow-up rather than a reason to rebase this.

Superseded → closing.

PR #332, opened 2026-06-20
nav: add "Studio" (rig_builder) to the LIBRARY sidebar group

byrongamatos:
> _🗂️ Migrated PR. Originally opened by @Jafz2001 on 2026-06-19 (was #533). Review history not preserved._

Pins the **rig_builder** plugin into the left sidebar's **LIBRARY** group as
**"Studio"**, between *Songs* and *Lessons* (faders icon). It's excluded from the
auto-built **PLUGINS** section so it doesn't appear twice; routing + topbar title
reuse the existing `plugin-rig_builder` screen path.

Small, self-contained change in `static/v3/shell.js`. Please review before merge.

### DIFF
--- a/static/v3/shell.js
+++ b/static/v3/shell.js
@@ -30,6 +30,7 @@
         { key: 'settings',   screen: 'settings',         label: 'Settings',        group: 'HOME',    icon: 'gear' },
         { key: 'playlists',  screen: 'v3-playlists',     label: 'Playlists',       group: 'LIBRARY', icon: 'list' },
         { key: 'songs',      screen: 'v3-songs',         label: 'Songs',           group: 'LIBRARY', icon: 'disc' },
+        { key: 'studio',     screen: 'plugin-rig_builder', label: 'Studio',        group: 'LIBRARY', icon: 'studio' },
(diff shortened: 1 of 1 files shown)


byrongamatos:
Closing as **superseded** — this feature is already on `main`.

The same idea (promote `rig_builder` to a first-class sidebar entry and take it out of the generic plugin list) landed via the nav rework in `d5e184d`. On `main` today:

- `rig_builder` is a dedicated nav entry (label **"Rig Builder"**, from the plugin's manifest `nav.label`), rendered **after** the LIBRARY group, gated on the plugin being installed.
- The generic per-plugin sidebar list was removed entirely (`renderPluginNav` → `renderPromotedNav`); the single **Plugins** entry is now the gallery's one entry point. SlopScale got the same first-class treatment under FeedBarcade.

Because `main` no longer has the `renderPluginNav` function or the NAV/ICONS layout this PR patches, it now shows as `CONFLICTING` and can't be merged as-is.

The one difference here was a naming/placement choice — **"Studio"** *inside* the LIBRARY group vs. **"Rig Builder"** *after* it. We kept "Rig Builder" (also avoids confusion with the separate `studio` plugin). If we want the "Studio" label or in-group placement, that's a one-line tweak on top of the current code rather than this branch.

Thanks @Jafz2001 for the original idea — it's effectively shipped. 🙏

PR #534, opened 2026-06-20
rename feedback refs

mogul:
- **fix: 'feed[dB]ack' -> 'fee[dB]ack' in v3 guitar tone source labels**
- **Update GitHub repo references from feedback* to feedBack***


ChrisBeWithYou:
**Queue note** — flagged from a full-queue integration dry-run.

This is one of ~30 near-identical `feedback*` → `feedBack*` rebrand PRs currently open across the org. They don't conflict with feature work, but each one drifts staler as its repo advances, so the conflict surface only grows the longer they sit — and a half-applied rebrand across repos is itself a plugin-loading risk.

**Recommend landing the whole rebrand as one coordinated batch** (or closing them if the rename is on hold) rather than piecemeal. Flagging so it's a deliberate call, not something that quietly rots in the queue.

Issue #31, opened 2026-06-20
MIDI record: Audible metronome click


byrongamatos:
> _Restored from `slopsmith/slopsmith#576` — original issue, opened by @byrongamatos on 2026-05-01._
> _[restored-from: slopsmith/slopsmith#576]_

Optional click track during recording (tick on each beat from the song's beat map). Volume slider; mute by default. Doesn't get baked into the song audio — purely a monitoring aid.

PR #554, opened 2026-06-22
fix(v3): promote Virtuoso to the first-class sidebar slot (was slopscale)

ChrisBeWithYou:
## What

Swap the practice plugin's first-class v3 sidebar promotion from the retired **SlopScale fork** (`id: slopscale`) to **Virtuoso** (`id: virtuoso`) in `static/v3/shell.js` — the NAV entry and the `PROMOTED_PLUGINS` slot.

## Why — companion to the bundle swap (cross-repo blocker)

The bundled practice plugin was rebranded/re-homed to [`got-feedback/feedback-plugin-virtuoso`](https://github.com/got-feedback/feedback-plugin-virtuoso) (`id: virtuoso`). The desktop bundle change is **got-feedback/feedBack-desktop#31** (`slopscale-fork` → `virtuoso`).

`shell.js` still promoted `slopscale`. `renderPromotedNav()` fills a promoted slot **only when that plugin id appears in `/api/plugins`**. After the bundle swap, `slopscale` no longer ships → no match → the dedicated sidebar slot silently goes dark and Virtuoso falls back to the generic **Plugins** gallery, losing its first-class entry.

This was flagged on feedBack-desktop#31 as a cross-repo blocker: *"Until that lands, this bundle change shouldn't ship to users or the practice plugin regresses in the UI."*

## Change

- NAV entry: `slopscale` → `virtuoso` (`screen: plugin-virtuoso`, label **"Virtuoso - Practice"** from the plugin's `nav.label`, same `target` icon).
- `PROMOTED_PLUGINS` slot: `navKey`/`pluginId` `slopscale` → `virtuoso`, `slotId` `v3-nav-slopscale` → `v3-nav-virtuoso`, **same `anchorAfter: 'feedbarcade'`** so the slot keeps SlopScale's exact placement (HOME group, right after FeedBarcade).
- Same mechanism as the editor promotion (#546). Gated on the plugin being installed, so it appears only when Virtuoso is loaded.
- CHANGELOG entry under `[Unreleased] → Changed`.

## Sequencing

Must land **with** the bundle swap (feedBack-desktop#31) — neither should ship to users alone, or the practice plugin regresses in the UI.


topkoa:
## Review

Core change is correct — verified `renderPromotedNav()` ([shell.js:266-275](https://github.com/got-feedback/feedBack/blob/main/static/v3/shell.js#L266-L275)) and the dark-slot reasoning in the description holds up. NAV + `PROMOTED_PLUGINS` swap is consistent: `screen: plugin-virtuoso` matches the `plugin-<id>` convention, `anchorAfter: 'feedbarcade'` preserves placement, gating unchanged. 👍

### Blocking: same bug class, missed on the Plugins gallery

`static/v3/plugins-page.js:103` still keys the dead id in its `CURATED` category map:

```js
flappy_bend: 'game', minigames: 'game', slopscale: 'game', chord_sprint: 'game',
```

After the bundle swap `slopscale` no longer ships, so this entry is dead. Category resolution for Virtuoso (`categoryOf`, line 121-122):

- manifest declares `category` → authoritative, fine
- manifest omits `category` → `CURATED['virtuoso']` is `undefined` → `deriveFromType` → **`'other'`** board

Unless `feedback-plugin-virtuoso`'s `plugin.json` sets `category`, the plugin drops to the generic "other" pedalboard on the v3 Plugins gallery — the exact "id no longer ships → surface regresses" failure this PR exists to prevent, just on the gallery instead of the sidebar.

**Please confirm the Virtuoso manifest sets `category`.** If it doesn't, add `slopscale: 'game'` → `virtuoso: '...'` here (and `'practice'` is arguably a better fit than `'game'` for a scale-practice plugin, though that's pre-existing).

### Non-blocking: doc drift

- `README.md:39` — plugin table still lists SlopScale + the old `slopsmith-plugin-slopscale` repo URL.
- `docs/plugin-capability-inventory.md:83` — `slopscale` row is now stale.

Worth folding into this PR or a fast follow.

### No action needed

- `tests/js/note_detection_domain.test.js:115` (`source: 'slopscale'`) — arbitrary fixture label, not plugin-id-bound.
- `data/progression/paths/*.json` "virtuoso" — coincidental skill-tier name, unrelated to the plugin.

### Sequencing

Confirmed feedBack-desktop#31 is still open. Agree these must land together — merging this alone while `slopscale` is still bundled leaves both ids mismatched and the slot dark. Good cross-repo callout in the description.

ChrisBeWithYou:
Thanks — good catch on the second surface. Addressed in `aa00be8`.

**Blocking (Plugins gallery):** confirmed — the Virtuoso manifest **does** set `"category": "practice"`, which `categoryOf()` treats as authoritative (shell.js path you cited; `plugins-page.js:121`), so it lands on the **practice** board, not `other`. No gallery regression. I still cleaned it up both ways:
- removed the dead `slopscale: 'game'` from `CURATED`, and
- added `virtuoso: 'practice'` (in the practice group) as a defensive fallback, so even a future manifest that omitted `category` wouldn't drop to `other`.

You're right that `'practice'` beats `'game'` here — Virtuoso is a practice studio, not a minigame.

**Non-blocking doc drift:** folded into this PR rather than a fast-follow:
- `README.md` — SlopScale row → Virtuoso (new `got-feedback/feedback-plugin-virtuoso` URL + clone + description).
- `docs/plugin-capability-inventory.md` — `slopscale` row → `virtuoso` (marked Active).

**No-action items:** agreed, left as-is — the `note_detection_domain.test.js` fixture label and the `data/progression/paths/*.json` "virtuoso" skill-tier name are unrelated to the plugin id.

**CI / sequencing:** the red checks here are the org Actions **billing** block ("job was not started… payments… spending limit"), not this diff — same as feedBack-desktop#31. And agreed, these two must merge together; shipping either alone leaves the ids mismatched.

ChrisBeWithYou:
@topkoa — ready for re-review when you have a sec. Your blocking item (the dead `slopscale` id in the Plugins-gallery `CURATED` map) is fixed in `aa00be8`, and I folded in the two doc-drift notes (README + plugin-capability-inventory). Details in my reply above. The red CI is just the org Actions billing block, not the diff. Still paired with feedBack-desktop#31 to merge together. Thanks!

PR #537, opened 2026-06-20
rename: slopsmith → feedBack, byron → got-feedBack

mogul:
## Summary

Renames all user-facing references from `slopsmith` to `feedBack` and `byron` to `got-feedBack` across the entire codebase.

### Changes

| Pattern | Replaced with |
|---|---|
| `slopsmith` | `feedBack` |
| `Slopsmith` | `FeedBack` |
| `SLOPSMITH` | `FEEDBACK` |
| `SlopSmith` | `FeedBack` |
| `byron` (name) | `got-feedBack` |
| `byron@ougsoft.com` | `hi@got-feedBack.org` |
| `/home/byron/` | `/opt/got-feedBack/` |
| `com.byron.` | `com.got-feedback.` |
| `github.com/byrongamatos/` | `github.com/got-feedback/` |

### Backward compatibility
- `window.slopsmith = window.feedback` alias kept for JS API compatibility
- `SLOPSMITH_*` env vars read with `FEEDBACK_*` fallback
- Protocol/storage strings migrated with read-old → write-new pattern

### Stats
- 248 files changed, 2239 insertions(+), 2236 deletions(-)
- Zero remaining `slopsmith` or `byron` references (excluding intentional backward-compat alias)

### Verification
- 7 rename scripts in `scripts/`
- `scripts/verify.sh` zero-tolerance check passes

PR #575, opened 2026-06-23
Edit region: Loop-in-3D round-trip between player and Song Editor

topkoa:
## What

Player-side half of the **Editor ⇄ 3D Highway region round-trip**. Lets you jump from the player into the Song Editor at whatever you're looking at, and lets the editor's "Loop in 3D" hand a region back to the highway as an A–B loop. The editor-side half is in **got-feedback/feedback-plugin-editor#32**.

## Highway → Editor

- **✎ Edit region** button in the loop controls (both v2 and v3) opens the Song Editor scrolled to the active A–B loop. With no loop set, it falls back to the section under the playhead, else a short window around it.
- **↩ Editor** button appears after a Loop-in-3D handoff to return to the exact edit position you came from.
- Both hidden unless the editor plugin is present (`typeof window.editSong`), state managed by `_updateEditRegionBtn`.

## Editor → Highway

- A one-shot `song:ready` listener consumes `window._pendingHighwayLoop` (set by the editor's "Loop in 3D" button) — after `playSong()`'s own `clearLoop()` has run — arming `setLoop(a,b)` over the region and auto-starting playback. Filename-guarded so a cancelled handoff can't arm a stale loop on an unrelated song.

## Notes

- Reuses the existing A/B loop API (`setLoop` / `getLoop`) — no new looping engine, no backend changes.
- Buttons added to both `static/index.html` (v2) and `static/v3/index.html` (separate file; v2 markup doesn't carry over), using only already-scanned Tailwind classes (no stylesheet rebuild).
- New globals `editRegionInEditor` / `returnToEditorFromHighway`; helpers `_resolveEditRegion` / `_updateEditRegionBtn`.

## Testing

Manual, both `/` (v2) and `/v3`: play a song, set Loop A/B (or rely on the section fallback), click ✎ Edit region → editor opens at that region. From the editor's ▶ Loop in 3D, confirm it returns to the same highway loop.

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -5764,7 +5764,7 @@ window.feedBack.on('song:ready', () => {
 // calls playSong(). Once the chart is ready (playSong's own clearLoop() has
 // already run, so the loop won't be wiped), arm the loop over the selected
 // region and start playback so the user lands inside the loop directly.
-window.slopsmith.on('song:ready', () => {
+window.feedBack.on('song:ready', () => {
     _updateEditRegionBtn();
     const pend = window._pendingHighwayLoop;
     if (!pend) return;
--- a/static/app.js
+++ b/static/app.js
@@ -5759,6 +5759,27 @@ window.feedBack.on('song:ready', () => {
     Promise.resolve(togglePlay()).catch((err) => console.warn('[app] autoplay failed:', err));
 });
 
+// Editor → Highway handoff (Editor ⇄ 3D Highway region round-trip). The
+// editor's "Loop in 3D" button stashes a pending loop + return context, then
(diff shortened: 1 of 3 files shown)


byrongamatos:
Thorough review done; verified it auto-merges cleanly into current `main` (no conflicts), all referenced globals/highway methods/CSS classes exist, `setLoop` seeks into the region, and the `!isPlaying` guard plays nicely with #537's autoplay. Rebased onto post-#537 `main` and pushed a one-line fix normalizing the new `song:ready` listener from the legacy `window.slopsmith` alias to canonical `window.feedBack` (head `5f745d2`). Merging together with the editor half (feedback-plugin-editor#32).

PR #585, opened 2026-06-23
Fix list/tree view: select mode, parts visibility, song actions

DeathlySin:
Brings the v3 list/tree view to parity with the grid card, and fixes a stale-CSS bug that made it (and #582's chips) render incorrectly on the Docker build.

## Frontend (`static/v3/songs.js`)
- **Select mode no longer collapses artist groups.** Expanding a group while in select mode wiped the whole tree on re-render, making selection unusable (the "circular" expand → collapse → can't-select bug). Expanded groups are now captured and restored across re-render.
- **Tree rows support multi-select** — per-row checkbox + selection ring, matching the grid card.
- **Clicking a row or arrangement chip in select mode now selects instead of playing** — the grid's capture-phase guard, applied to the tree (`<summary>` headers sit outside `[data-fn]`, so native expand/collapse is untouched).
- **Favourite / Save for later / overflow menu (⋮) are always visible** on tree rows (not hover-only), matching the grid card.

Builds on #582's `arrChipsHtml` tree chips — no duplication, just fills the remaining gaps.

## The Docker fix (`static/tailwind.min.css`)
The committed Tailwind CSS was **stale**: `.sm\:flex` (and the other new utilities) were never compiled in, so #582's `hidden sm:flex` arrangement chips — and the new action cluster — rendered `display:none` on the **Docker** build, which serves the committed CSS as-is. The **Desktop** build looked fine only because it rebuilds Tailwind from source at bundle time. Regenerated `static/tailwind.min.css` with the pinned `tailwindcss@3.4.19` so Docker matches Desktop. This also restores #582's chips for every Docker deploy.

## Testing
- **Docker:** driven headless — select/deselect persists across group expand/collapse, chip clicks toggle selection instead of playing, save-for-later toggles, non-select playback unaffected, and chips/badges/action cluster are all visible.
- **Desktop:** verified the updated assets render correctly in the installed build.


DeathlySin:
Leave this for now, there's something gone wrong somewhere after testing it again. I'll reopen when I've figured out what's wrong.

PR #589, opened 2026-06-24
feat(achievements): wall sync drain worker (epic PR3, client side)

byrongamatos:
**Achievements & Feats of Power epic — PR 3 of 3 (client side).** Stacked on the PR2 branch. Wires the dead-letter wall-sync worker.

The bundled `achievements` plugin gains a background worker that POSTs queued Feat unlocks/removals to the hosted **feedback-achievements** wall (separate repo). Idle unless `FEEDBACK_ACHIEVEMENTS_WALL_URL` is set; uses `requests` + the client-token header (mirrors `lib/lyrics_transcribe`).

**Dead-letter, never drop** (pure `engine.drain_decision`):
- network error / `429` / `5xx` → keep `pending` (retry)
- other `4xx` → `dead_letter` (diagnosable, replayable)
- `2xx` → delete on server ack

`remove-me` now enqueues a wall removal keyed by the reused `player_hash`.

### Verification
- `tests/plugins/achievements/test_sync.py`: decision table + ack/retry/dead-letter retention + four-field on-the-wire payload.
- Native **end-to-end staging round-trip** against a local wall: earn a Feat → drains onto the wall (name + short hash) → `remove-me` → wall empties — **no IP** in tables or access logs.

The hosted service itself (FastAPI + SQLite, Feats-only, hidden-until-first-global-unlock, profanity/rate-limit/token, `render.yaml`, `migrate.py`) lives in the new **`feedback-achievements`** repo (pushed separately).

> Final epic PR. Org CI may be billing-blocked — verified locally. **Do not merge without byron's sign-off.**


byrongamatos:
## Local review pass (recall-biased, 5 finder angles × 2 repos)

Ran a thorough local review across both repos before sign-off. **6 real bugs + 3 cleanups fixed** (commit `0e95d76` here; wall-side in `feedback-achievements@904e084`). All verified locally — 45 plugin tests + 12 wall tests, Playwright smoke, native round-trip.

### Fixed — correctness
1. **`secret_witching` Feat was dead** — `post_activity` wrote `witching_nights_run` to the DB *before* snapshotting `prev_tiers`, so `diff_unlocks` never saw the fresh unlock. Folded the run into the activity delta (the asymmetry `chart_encore` already relies on). +2 regression tests.
2. **`chart_encore` broke across restarts** — per-chart key used `abs(hash(str))`, which Python salts per process (`PYTHONHASHSEED`), so the same chart hit a new counter each restart and could never reach 100/500. Now a stable sha1 digest. +regression test.
3. **`note:hit`/`miss` had no active-song gate** — tuner/input-calibration note events could inflate Feats and flush a phantom `chart:null` session. Gated on an `active` flag (set on `song:loading`, cleared on `song:ended`/`song:stop`).
4. **Wall cache invalidation race** — `_wall_cache` was read/written without a lock vs FastAPI's threadpool; a build that read the DB just before the first unlock could pin a stale empty wall for the full TTL (defeating hidden-until-first-unlock). Now version-gated under `_lock`.
5. **Wall `_rate` unbounded growth** — one permanent `(hash, ip)` entry per pair (spoofable via `X-Forwarded-For`) → slow leak on the pinned single instance. Now sweeps expired buckets past a threshold.
6. **`get_feats` could 500** — `f["id"]` on a malformed catalogue entry. Skips entries without an id.

### Fixed — cleanup / convention
7. `_read_counters` no longer pulls the unbounded `chart_plays:*` rows on every activity POST.
8. **P-III**: plugin localStorage key prefixed → `achievements:profile-cat`.
9. Extracted the duplicated local-ISO-date helper in `screen.js`.

### Noted / accepted (not fixed — rationale)
- `_record_unlock` equal-tier early-return won't re-class an id from competency→feat at the same tier — narrow; the baseline never does this.
- `evaluateBaseline` re-POSTs already-earned tiers on progression bursts — idempotent server-side, low volume.
- Drain worker opens a connection per queued row; `_opted_in` re-parses `config.json` per unlock — minor, low volume.
- **Pre-existing** onboarding race (paths-step submit enabled before tiles load) — not introduced by this epic; the step renumber slightly widens it. Flagging for a separate fix.
- Altitude nits: witching 2–5am window + `INSTRUMENTS` fallback are hardcoded in JS; `renderFeats` leans on core-guaranteed Tailwind classes. Follow-ups.
- `personal_best` / `challenger` baseline criteria remain best-effort stubs (known follow-up).

Still needs your sign-off; org CI may be billing-blocked so everything was verified locally.

PR #611, opened 2026-06-27
fix(player): keep play/pause button in sync when a JUCE reroute aborts autoplay's play()

ChrisBeWithYou:
## Bug
On the **first song after a fresh load** on desktop, the play/pause button shows ▶ (Play) while the song is actually playing — so it takes **two clicks to pause** (one to resync the flag, one to actually pause). Reported on Mac, reproduced on Windows; only the first song after a hard reload, both via the desktop app.

## Root cause
On a fresh load the JUCE audio engine is usually still starting when the first song loads, so the song begins on the HTML5 `<audio>` element (`_juceMode = false`). The engine-reroute watcher then migrates it HTML5→JUCE, and its first step is a deliberate `audio.pause()`. That **aborts autoplay's in-flight `togglePlay()` `audio.play()`** (AbortError) — even though playback continues on the JUCE transport. `togglePlay()`'s catch then set `isPlaying = false` and the button to ▶, while JUCE kept playing.

The reroute already guards the `<audio>` `play`/`pause` **DOM listeners** with `window._juceRerouteInProgress` so its transparent pause doesn't flip plugin state — but `togglePlay()`'s own catch wasn't guarded.

## Fix
Extend the same guard to `togglePlay()`'s catch and the count-in catch: a `play()` rejection caused by the reroute's deliberate pause is not a real failure, so don't reset the button. A genuine failure (outside a reroute) still resets correctly.

## Test
`tests/js/play_button_reroute_guard.test.js` extracts `togglePlay()` and drives it through a reroute-aborted `play()`, asserting the button stays ⏸ and `isPlaying` stays true; a control case proves a genuine rejection still resets to ▶. The regression test **fails without the guard**, passes with it. All 56 related JS tests pass.

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -6270,6 +6270,14 @@ async function togglePlay() {
         } catch (err) {
             if (sessionGen !== _audioSeekGen) return;
             if (attempt !== _playAttemptGen) return;
+            // An engine reroute (HTML5 -> JUCE) deliberately pauses the <audio>
+            // element mid-migration, which rejects this in-flight play() with an
+            // AbortError even though playback continues on the JUCE transport.
+            // The reroute owns isPlaying / the button while it runs (same guard
+            // the <audio> 'play'/'pause' listeners use); resetting here would
+            // leave the button showing Play while the song keeps playing — the
+            // "two clicks to pause on the first song after a fresh load" bug.
+            if (window._juceRerouteInProgress) return;
             console.error('[app] audio.play() rejected:', err);
             isPlaying = false;
             setPlayButtonState(false);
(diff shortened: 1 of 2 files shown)


byrongamatos:
Reviewed (manual + local `codex review`) — **LGTM, no issues found.**

The `if (window._juceRerouteInProgress) return;` guard correctly extends the existing refcounted flag (set/decremented around the reroute at app.js:4804/4911/4932/4940, already guarding the `<audio>` play/pause listeners at ~5655/5679/5689), so it's not dead code and won't get stuck (decrement is `Math.max(0, …)` in the completion path). The added regression test covers both the reroute-abort case (button stays Pause) and a genuine rejection (button resets to Play). Codex independently found no introduced bug.

PR #619, opened 2026-06-27
fix(player): reliable Escape "Back" + resumable, optionally-confirmed song exit

ChrisBeWithYou:
## Why

Pressing **Escape didn't reliably leave a song**. Clicking any transport control (Play / FF / RW / Restart) leaves that `<button>` focused, and `_shortcutDispatchBlocked()` bails the shortcut dispatcher for any focused `INPUT/SELECT/TEXTAREA/BUTTON` — so the player-scope `Escape = Back` shortcut never fired until you clicked empty canvas to blur the control. **Space** already had a player-screen carve-out (#593); Escape didn't. That asymmetry was the bug.

Reported in Discord (MajorMokoto): *"escape … does not always exit the song. If you've clicked on any of the controls … escape will not do anything till you click on the main window."* The thread also asked that Escape behave like a universal **Back**, and floated an optional "are you sure?" guard (with **double-Escape** / **Space-to-confirm**).

Scoped via a multi-seat design charrette (engagement-psychology, learning-design, operability, codebase-reality). Consensus: make leaving a song **reliable and recoverable, not gated** — the confirm is opt-in only.

### Phase 1 — focus fix (the bug)
Generalize the Space carve-out in `_shortcutDispatchBlocked` to Escape, on the **player and settings** screens (both register `Escape = Back`; settings had the identical latent bug). The earlier guards still win first: text inputs (Esc clears/blurs the field), the Section Practice popover (already claims Esc), and a true modal (`role="dialog" aria-modal="true"` / `.feedBack-modal`) still traps Esc rather than ejecting past it. Plugins' player-scope Escape shortcuts are fixed identically — no documented key changes.

### Phase 2 — resume where you left off
Leaving the player snapshots `{song, arrangement, position, speed}` to `localStorage`; a non-blocking **"Resume practice"** pill offers it back on the next non-player screen / next launch. `playSong()` gains a `{resume}` option that restores speed + seeks to the saved position on `song:ready` instead of the normal autostart. Conservative (ignores < 3 s / near-end), cleared on natural song-end and once consumed, expires after 24 h. Public: `window.resumeLastSession()`. (The deeper nav-state work — returning to a song after wandering into Settings → Tone Builder — is a separate, larger track.)

### Phase 3 — opt-in "Ask before leaving a song" (Gameplay tab, default OFF)
A **true-modal** confirm with **Escape = Stay** (dismiss the prompt, like every other modal — a second Escape does *not* leave) and **Space / Enter / the focused “Leave” button = leave** (Leave is default-focused). The player Escape shortcut and the v3 ✕ route through `window.requestExitSong()`; auto-exit on song-end and a results screen's own Close stay unguarded.

## ⚠️ Testing status — needs a smoke test

**Automated Playwright specs were written but NOT executed locally** — the Playwright Chromium browser install wedged on extraction in this environment (download/unzip, unrelated to the code). The 16 new specs are committed and mirror the repo's existing passing patterns:

- `tests/browser/keyboard-shortcuts.spec.ts` — **5 Escape cases**: focused-button repro, text-input no-exit, no-escape-past-modal, Section Practice popover, settings twin-bug (mirror the #593 Space tests).
- `tests/browser/resume-session.spec.ts` — **6 cases**: snapshot guards, staleness, pill show/hide/dismiss, resume consumption.
- `tests/browser/exit-confirm.spec.ts` — **5 cases**: default-off instant exit, confirm-on opens + stays, 2nd-Escape dismisses (Stay), Stay/Leave, Enter-leaves.

### @byrongamatos — flagged for your smoke test tomorrow 🙏
1. In a song, click **Play** (or FF/RW/Restart) then press **Escape** → should exit immediately (this is the reported bug).
2. Re-enter a song, exit, confirm the **"Resume practice"** pill appears and resumes at the right position / speed / arrangement.
3. Settings → Gameplay → **"Ask before leaving a song"** ON → Escape opens the confirm; **Space/Enter (or Leave)** leaves while a **second Escape dismisses it (Stay)**; **Stay** keeps you in.
4. Default (toggle OFF) → Escape still leaves instantly.

Code is syntax-verified; Phase 1 mirrors the proven #593 carve-out.

https://claude.ai/code/session_01QbexxfTt8q2tAn436MqGWF


ChrisBeWithYou:
## ✅ Local Chromium run — new behavior is green

Ran the browser specs against a local server (Chromium via Playwright). Updated the testing status from the original PR body:

**All 16 new specs pass** + the existing #593 Space tests:
- `keyboard-shortcuts.spec.ts` — 5 Escape cases ✓
- `resume-session.spec.ts` — 6 cases ✓
- `exit-confirm.spec.ts` — 5 cases ✓

**No regressions from this PR.** 4 unrelated tests fail in this bare environment (`should show library shortcuts in help modal`, `focus library search for /`, `Shift+Slash while typing`, `viz picker focused`) — all `?`/`/`/search/viz-picker tests this diff never touches. I confirmed they **fail identically on unmodified `main`** (b206633, real library on :8000), so they're pre-existing/environmental, not introduced here.

Follow-up commit `7ba45e3` hardens the three specs (stub `GET /api/profile` to skip the first-run onboarding overlay, which is a modal that was covering the player) so they run cleanly from a fresh profile.

**@Byron** — manual smoke test still appreciated for the real feel (resume seek accuracy, confirm-modal UX), but the automated behavior is now verified green.

ChrisBeWithYou:
## Refinement charrette (round 2) — synthesis

After this PR's first cut shipped (Escape=Back focus fix + resumable exit + opt-in confirm) and Byron smoke-tested it ("working FANTASTIC"), he gave four refinement asks. We ran a 6-seat design panel — gamification, learning-design, UX, UI, marketing, dev-ops. The amend on this branch (`Escape cancels back to song, pause on open/resume on stay`) implements the core decisions below; item **#4** lands separately in **note_detect [#44](https://github.com/got-feedback/feedback-plugin-notedetect/pull/44)**.

### The throughline all six converged on

Escape should mean one thing everywhere — **"close the frontmost layer, choosing the safe option"** — and friction belongs **only where real loss can happen**. Exit is already resumable (nothing lost) → keep the doorway frictionless. The **results card is the one place loss is real** (your score) → that's the one place to require a deliberate choice. *"Protect the trophy, not the doorway."*

### Decisions

| # | Question | Decision | Rationale | Where |
|---|----------|----------|-----------|-------|
| 1 | Confirm default ON/OFF | **Stays OFF** | A confirm signals "you'll lose something," but exit is resumable so it guards nothing and nags a high-frequency action. Tester wanted ON; declined (one-line flip if reconsidered). The "protect me" instinct is met by #4 instead. | this PR (unchanged) |
| 2 | Double-Escape on the open prompt | **Cancel back to the song** (was: 2nd Esc leaves) | Esc = dismiss-the-frontmost-modal everywhere (matches the app's own `_confirmDialog`). Only opted-in users see the prompt. Tester **retracted** his earlier "double-esc leaves." Leaving stays the explicit, default-focused **Leave** button → Space/Enter/click = "just get me out." | this PR |
| 3 | Pause on prompt / resume on cancel | **Yes** | Don't run/score the song behind the prompt. Resume **only what we paused** (wasPlaying), only if it's still the same live song on the player; cancel any count-in; never auto-resume a song the user had paused. Via the canonical `togglePlay()` path. | this PR |
| 4 | Results card: outside-click leaves | **Remove it** | A terminal scorecard shouldn't vanish on a stray click — explicit Exit/Retry/Return only. | note_detect #44 |

L&D dissented on #2 (wanted monotonic "Escape always leaves") — absorbed: with the confirm **OFF** (the default), a single Escape leaves instantly, so the monotonic-Back reflex is fully preserved for power users; #2 only governs the prompt a user opted into.

### ⚠️ Verification note for this PR

The pause/resume (#3) adds a new pause→resume cycle on the **desktop JUCE** transport (the known play/pause-desync path). The Playwright mock has no backing audio, so the specs lock the navigation + keyboard semantics but **the audio pause/resume itself needs a desktop-build smoke-test before merge.**

### Deliberately NOT done

- Default ON (declined — see #1).
- "Discard unsaved changes?" copy / dwell timers on the results card (dark-pattern-adjacent).
- Escape=Done keybinding on the results card (panel suggested it; out of scope for the reported backdrop-click bug — possible follow-up).

PR #620, opened 2026-06-28
fix(v3): refresh song/lesson score badge on first return (decode stats:recorded filename)

ChrisBeWithYou:
## Problem

Reported in testing (0.3.0 / 0.2.9 beta): after finishing a song or lesson, its **accuracy badge** on the Songs screen stays stale — it only updates after an app **restart** (or, partially, after a search / navigating away and back). [#574](https://github.com/got-feedback/feedback/pull/574) added the intended mechanism (`stats-recorder` emits `stats:recorded` after the score POST resolves; `songs.js` repaints the badge in place), but the badge still didn't refresh on the first return from a song.

## Root cause

A filename **key-space mismatch** — the in-place repaint never matched a card, so it silently no-oped:

- `playCard` calls `playSong(encodeURIComponent(localFilename))`, and `playSong` emits `song:loading` / (downstream) `stats:recorded` with that **URL-encoded** filename (e.g. `Black%20Me%20Out.sloppak`).
- Library cards key on the **decoded** `localFilename` (`data-fn="Black Me Out.sloppak"`), and `/api/stats/best` is server-canonicalized to that same decoded key (`server.py` `_canonical_song_filename`).
- So `repaintAccuracy`'s `el.getAttribute('data-fn') !== key` check rejected **every** card, and `state.accuracy[encoded]` was `undefined`.

Only a full `render()` re-derives each badge from the card's own decoded filename, which is why restart / search / re-enter "worked" but the first-return in-place repaint didn't. Every affected song name in the bug report has a space, so it failed for all of them.

## Fix

`static/v3/songs.js` — a small `decFn` helper decodes the `stats:recorded` filename back into the card / `state.accuracy` key space before marking dirty and repainting. This fixes **both** the immediate-repaint path (library visible) and the `onV3SongsScreenEnter` deferred path (marked dirty, applied on return).

`decFn` is idempotent for already-decoded names (no `%` → returned as-is) and falls back to the original on malformed input (`try/catch` around `decodeURIComponent`), so a real filename containing a literal `%` is never corrupted.

## Testing

- **New:** `tests/js/v3_songs_score_badge_refresh.test.js` extracts the **real** `decFn` from the shipped source via `vm` (not a mirror) and proves the encoded event filename round-trips back to the raw card key — covering spaces, a subfolder `/` (→ `%2F`), idempotence, a literal-`%` filename (no throw), and non-string input. 4 cases, all green.
- Full `tests/js` suite: this change adds 4 passing tests and breaks nothing. (The suite has 12 pre-existing, unrelated failures — `highway_colors_facade`, `loop_api`, `capabilities_manifest_runtime`, `legacy_shim_hits`, `live_guitar_tone_source` — confirmed identical on pristine `origin/main`.)

Follow-up to #574 (which is merged, hence a new PR rather than an amend).

https://claude.ai/code/session_01QbexxfTt8q2tAn436MqGWF


ChrisBeWithYou:
### Investigation summary (songs fix here; lessons checked separately)

The original report flagged stale scores on **both** the song menu and the lessons menu. They turned out to have separate data paths — only the songs path was actually broken, which is what this PR fixes.

**Songs (this PR) — real bug.** PR #574 wired up `stats:recorded → repaintAccuracy`, but the repaint never matched a card. `playCard` launches via `playSong(encodeURIComponent(localFilename))`, so the event carries the **URL-encoded** filename, while cards key on the **decoded** `localFilename` (`data-fn`) and `/api/stats/best` is server-canonicalized to that same decoded key (`server.py:_canonical_song_filename`). So `data-fn !== key` rejected every card and `state.accuracy[encoded]` was `undefined`. Only a full `render()` (restart / search / re-enter) re-derived badges from each card's own decoded filename — exactly the "only shows after a restart" report. `decFn` decodes the event filename back into the card key space, fixing both the immediate and the `onV3SongsScreenEnter` deferred repaint.

**Lessons — checked, no shared bug, not touched here.** `static/v3/lessons.js` uses a different, self-healing path: `render()` unconditionally re-fetches `/api/plugins/tutorials/progress` (the per-lesson `best_accuracy`) on every `screen:changed → v3-lessons`, plus on `progression:updated` / `v3:profile-updated` — no filename key, no encode/decode mismatch, and no cached-DOM fast-path. And in the `tutorials` plugin (v1) the score is self-reported: the `best_accuracy` write is an **awaited** `POST /api/plugins/tutorials/runs` done while the user is still on the `plugin-tutorials` screen, so it's committed before the catalog is ever re-shown. No analogous race or key mismatch to fix.

**Net: one PR, not two.** Worth re-testing both screens once this merges; if lesson scores still look stale on first return afterward, that'd be a lessons-specific timing thing (e.g. core auto-exit vs. the plugin's record flow), not this root cause.

PR #647, opened 2026-06-29
perf(tuner): idle always-on tuner viz rAF when there's no signal

byrongamatos:
## Why

Running the v3 desktop AppImage pins a renderer core even while idle on the **home screen** doing nothing. The home tuner card is intentionally always-on, but every tuner visualization drove a **self-rescheduling 60fps `requestAnimationFrame` loop that never stopped** — repainting a static needle/strobe forever.

(Diagnosed alongside the native side: the engine's ML detection pipeline also runs unconditionally to feed the always-on tuner, which only needs raw frames — that's a separate desktop PR. This PR is the renderer half.)

## What

Each viz now idles its rAF once there's nothing left to animate, and re-kicks from `update()` on the next reading that actually moves it:

- **analogue-gauge** — stop when the needle + drum strip have settled on their targets (`|target − current|` below a sub-visible epsilon); restart when a new reading moves the target.
- **strobe / mace-fx-iii / chef-mt3** — stop when there's no live signal and the strobe drift (and glow fade) have fully decayed; restart on the next note.
- **toilet-tuner** — stop when silent and the plunger has eased back to centre; `_kick()` is guarded so repeated no-signal `update(null)` calls don't re-kick a parked loop.

Active tuning is unchanged: the loop runs whenever a note is sounding or the indicator is still moving. Net effect — a silent home tuner stops repainting instead of spinning at 60fps.

## Verification

- `node --check` passes on all five viz files.
- All `requestAnimationFrame` usage in the tuner plugin is confined to these viz files (grep-confirmed) — no other always-on renderer loop.
- Live CPU reduction should be confirmed in a fresh bundle/AppImage.

Bumps tuner `1.3.1 → 1.3.2`.

### DIFF
--- a/plugins/tuner/plugin.json
+++ b/plugins/tuner/plugin.json
@@ -1,7 +1,7 @@
 {
   "id": "tuner",
   "name": "Guitar/Bass Tuner",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "bundled": true,
   "private": false,
   "script": "screen.js",
--- a/plugins/tuner/visualization/analogue-gauge.js
+++ b/plugins/tuner/visualization/analogue-gauge.js
@@ -18,6 +18,8 @@
     // ── Constants ─────────────────────────────────────────────────────
     var _TUNER_LABEL_H = 12;           // px height of each drum label
     var _TUNER_NEEDLE_HALF_SWEEP = 90; // degrees — ±50 cents = horizontal (180° apart)
+    var _SETTLE_A = 0.05;              // deg — needle "settled" threshold (sub-visible)
+    var _SETTLE_Y = 0.1;               // px — drum-strip "settled" threshold
     var _TUNER_IN_TUNE_THRESHOLD = 2;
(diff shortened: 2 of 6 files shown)


byrongamatos:
**Codex (gpt-5.4) review:** no correctness regressions — confirmed no stuck/frozen-indicator restart bug, no first-frame `dt` jump, no kick-every-frame defeat (incl. toilet-tuner's per-update `_kick` guard), and `destroy()` still safe.

Its three findings are all the *same secondary opportunity*: the loops could also idle during a **sustained live note** (e.g. a held in-tune reading where the strobe/needle has stopped moving), not just on no-signal. That's deliberately out of scope here — this PR targets the **reported** symptom (idle home screen, *no signal*, loops spinning at 60fps), which is fully addressed. Idling during active sustained tuning is riskier (kick-on-every-signal-update interplay + freeze-during-live concerns) for a marginal active-use-only gain; captured as a possible follow-up.

PR #651, opened 2026-06-29
fix(input_setup): keep backdrop-blur during calibration wizard

OmikronApex:
## Summary

- Hiding the entire `#input-setup-overlay` removed it from the render tree, stripping its `backdrop-blur-sm` — so the background blur vanished while the calibration wizard was open and returned when it closed.
- Fix: hide only the `[data-is-host]` card inside the overlay; the overlay element itself stays in the DOM so its backdrop-blur remains active throughout.

## Test plan

- [ ] Open initial setup wizard
- [ ] Verify background blur is visible before starting calibration
- [ ] Click the calibration button — confirm blur persists while calibration wizard is open
- [ ] Complete or cancel calibration — confirm input-setup card is restored normally

### DIFF
--- a/plugins/input_setup/screen.js
+++ b/plugins/input_setup/screen.js
@@ -164,20 +164,21 @@
 
                 host.querySelector('[data-is-cal]').addEventListener('click', () => {
                     if (hasDetector) {
-                        // Hide our own full-screen overlay while note_detect's
-                        // Calibration Wizard runs on top. That wizard goes
-                        // transparent (pointer-events:none) when it minimizes to
-                        // expose the Tuner; if our overlay stayed up it would show
-                        // through — covering the tuner with the still-mounted
-                        // "select your input" card and looking like a second
-                        // wizard at the input step. Restore it on done/cancel
-                        // (one of which always fires when that wizard closes).
+                        // Hide only the inner card while note_detect's Calibration
+                        // Wizard runs on top — keep the overlay element in the DOM
+                        // so its backdrop-blur stays active. The wizard goes
+                        // pointer-events:none when it minimizes to expose the Tuner;
+                        // hiding just the card prevents the "select your input" panel
+                        // from showing through without removing the background blur.
+                        // Restore on done/cancel (one always fires when wizard closes).
                         const ov = document.getElementById('input-setup-overlay');
-                        const prevDisplay = ov ? ov.style.display : '';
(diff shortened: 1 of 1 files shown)


byrongamatos (changes requested):
Thanks for chasing the lost blur — that's a real glitch. But I think this approach regresses the tuner exposure that the original full-overlay-hide was protecting (the failure mode the `screen.js:167` comment calls out, and what core #577 fixed).

## The problem

The change keeps `#input-setup-overlay` mounted (`fixed inset-0 z-[210] bg-black/60 backdrop-blur-sm`, **pointer-events:auto**) and hides only the inner card. But that overlay is full-screen, and the layering is:

| Layer | z-index |
|---|---|
| note_detect calibration wizard | **300** |
| **#input-setup-overlay (kept by this PR)** | **210** |
| player / highway / transport / tuner | **1–10** |

When the user hits **"Minimize Wizard — Play Song"**, the wizard (z-300) sets *itself* `background:transparent; pointer-events:none` to expose the player/tuner underneath. With the overlay still mounted, the z-210 layer now sits **between** the transparent wizard and the player — so during the minimized state the tuner/player is:

- **dimmed** (`bg-black/60`),
- **blurred** (`backdrop-blur-sm`),
- and **click-blocked** — clicks fall through the now-`pointer-events:none` wizard and are swallowed by the z-210 overlay before they reach the player controls (z 1–10).

So "Minimize Wizard — Play Song" (and any tuner interaction during minimize) breaks. The original code hid the *whole* overlay precisely to avoid this. None of the test-plan steps here exercise the minimized state, which is where it shows up.

## Suggested fix

Move the `backdrop-blur-sm` onto the **calibration wizard overlay itself** (note_detect side, z-300) instead of relying on input_setup's overlay. Then:

- wizard shown → blur present (on the wizard, above everything);
- wizard minimized → it goes transparent, so the blur correctly vanishes and the player/tuner is sharp + interactive;
- input_setup keeps fully hiding its overlay → no occlusion.

That gets the blur you want without re-occluding the tuner — but it's a change in note_detect, not here. Alternatively, if you want to keep it in input_setup, the overlay would at minimum need `pointer-events:none` **and** to drop its dim/blur whenever the wizard is in its minimized state — which is more coupling than it's worth.

Happy to take the wizard-side approach if you'd prefer.

PR #652, opened 2026-06-30
Aspect-aware framing for ultra-wide 3D highway panes

topkoa:
## What

On a **top/bottom 2-player split**, each 3D highway pane is full-width / half-height (~32:9). The camera's vertical FOV was fixed, so at that aspect the horizontal cone ballooned past ~130° and crushed the fixed-width neck into a thin central sliver with big dead margins.

This adds a **horizontal-FOV-hold** ("Hor+") path: past a configurable *start aspect*, the effective vertical FOV is lowered so the horizontal cone stays ~constant — the neck fills a wide pane instead of collapsing. At/under the start aspect it's an **exact no-op**, so normal ~16:9 single-player and most 2×2 panes render unchanged. Optional pose nudges (height / dolly / pitch / look-depth) flatten the view toward a low, immersive angle.

**Shipped on by default for wide panes** (panes wider than ~2.25:1) so testers see it immediately. Toggling it off restores the exact prior framing, so it doubles as an A/B control.

## How to use / test

1. Open a song, enable a **top/bottom 2-player split** (P1 + P2, both **3D Highway**). This is where it matters most.
2. The new framing is **on by default** on wide panes — the neck should fill the pane width.
3. Press **`Shift+A`** (player focused) to toggle the feature **A/B**:
   - **OFF** = the previous framing (baseline).
   - **ON** = the new wide-pane framing.
4. With it on, a **tuner panel** appears (top-right). Drag sliders to adjust live; both panes update every frame:
   - **Base vFOV°** — base vertical field of view.
   - **Start aspect** — panes wider than this get the effect; narrower are untouched.
   - **Blend** — 0 = off, 1 = full hold.
   - **Min vFOV°** — floor on how far the FOV can drop.
   - **Height × / Dolly × / Pitch +** — camera pose nudges.
   - **Look depth** — aims the camera further down the neck; flattens the mid-distance "hump".
   - **Override held hFOV°** — pin the held horizontal cone explicitly.
   - Live readout shows the current **pane aspect · effective vFOV**.
   - **Reset** → back to the shipped defaults. **Copy** → dumps the current values as JSON to the console + clipboard (paste into a comment here to suggest a tuning).
5. Settings **persist to `localStorage`** (`h3d_aspect_tune`) and survive reloads. To wipe back to defaults: `localStorage.removeItem('h3d_aspect_tune')` then reload.

### What to check
- **Single-player** and **2×2** (each pane ~16:9) look **unchanged** with the feature on.
- **Top/bottom** panes now fill the width; the neck reads clearly end-to-end.
- `Shift+A` cleanly flips between old and new in a live split.
- Tuner panel readout/sliders track the render.

## Tests
`tests/js/highway_3d_wide_fov.test.js` source-pins the constants, the no-op-at-start-aspect guarantee, the change-guarded FOV write, the persisted bridge, the panel, and the lifecycle reset. Run:
```
node --test tests/js/highway_3d_*.test.js
```

## Notes / open questions
- **Default-on for wide panes** is deliberate for testing — open to flipping it default-off and gating behind a setting before merge.
- The shipped tuning is an in-progress pick; final values are easy to swap (single `_ASPECT_DEFAULTS` object). Suggestions welcome via the panel's **Copy** JSON.

closer-Screenshot 2026-06-29 200906

### DIFF
--- a/plugins/highway_3d/screen.js
+++ b/plugins/highway_3d/screen.js
@@ -1648,17 +1648,22 @@
     // a tuning session survives reloads; the floating panel (Shift+A) writes the
     // same object live. All of this is a debug aid — none of it runs unless the
     // user opts in.
-    const _ASPECT_LS = 'h3d_aspect_tune';
-    // Working defaults — the current wide-pane tune, shipped ON for testing.
-    // With enabled:true the framing engages on panes wider than startAspect
-    // (2.25); normal ~16:9 panes (single-player, most 2x2 layouts) are below
-    // that and render unchanged. Toggling OFF (Shift+A) restores the exact prior
-    // framing — effectiveVfov returns the geometric base (BASE_VFOV) and the
-    // pose nudges gate off — so it doubles as the A/B baseline. Reset lands here.
-    // localStorage overrides all of this per machine.
+    // Versioned key: the first iteration shipped a broken default (enabled:true,
+    // baseVfov:30) and may have persisted it. Bumping the key ignores that stale
+    // state so the corrected default-off config actually takes effect.
+    const _ASPECT_LS = 'h3d_aspect_tune2';
+    // Working defaults. Default OFF, so out of the box this is an exact no-op —
+    // every pane renders byte-for-byte as before (effectiveVfov returns
+    // BASE_VFOV and the pose nudges gate off). The config is also coherent when
+    // a tester turns it ON via Shift+A: baseVfov == BASE_VFOV so normal ~16:9
+    // panes (single-player, most 2x2) stay at 70° even enabled, and only panes
+    // wider than startAspect (2.25) engage the Hor+ hold; blend:1 makes that
+    // hold actually take effect; minVfovDeg (28) sits below baseVfov so the floor
+    // is a real floor. The pose nudges are the in-progress wide-pane look a
+    // tester sees once enabled. localStorage overrides all of this per machine.
     const _ASPECT_DEFAULTS = {
-        enabled: true, baseVfov: 30, startAspect: 2.25, hfovDeg: null,
-        blend: 0, minVfovDeg: 36, splitOnly: false,
+        enabled: false, baseVfov: BASE_VFOV, startAspect: 2.25, hfovDeg: null,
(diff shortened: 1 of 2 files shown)


byrongamatos:
Pushed a fix commit (3aeeac2) addressing the blocking issues — the algorithm and tuning panel are untouched, just the shipped config.

**What was wrong:** `_ASPECT_DEFAULTS` shipped `enabled:true, baseVfov:30, blend:0, minVfovDeg:36`, which contradicted the "default off / normal panes unchanged" contract:
- `enabled:true` + `baseVfov:30` forced **every** pane's vFOV from 70° → 30° (single-player/2×2 included — a global zoom, not a no-op).
- `blend:0` collapsed the Hor+ math back to base, so the actual horizontal-hold did nothing even on wide panes.
- `minVfovDeg:36 > baseVfov:30` was an inverted floor (clamped wide panes *up* to 36°).

**New defaults:** `{ enabled:false, baseVfov:BASE_VFOV(70), blend:1, minVfovDeg:HORPLUS_MIN_VFOV(28) }`:
- **OFF by default** → `camUpdate` passes a null tune → `effectiveVfov` returns `BASE_VFOV` → exact no-op on every pane (verified 70° at 16:9 *and* 32:9).
- **When enabled (Shift+A)**: `baseVfov==BASE_VFOV` keeps normal/≤start panes at 70°, and `blend:1` makes the hold actually engage on wide panes (47.7° at 32:9, flooring toward 28° as it widens). So the A/B now compares *real* framing vs baseline instead of zoom vs baseline.

Also bumped the localStorage key (`h3d_aspect_tune` → `h3d_aspect_tune2`) so machines that persisted the old broken default pick up the corrected one, and added source-pin tests guarding default-off + the coherent base/blend/floor. 110/110 pass.

Left your pose-nudge values (`heightMul:0.30` etc.) as-is — that's the in-progress look, now dormant until enabled. Two minor things I did **not** change, since they're intentional session-global dev aids and harmless now that the feature is off by default: the `Shift+A` shortcut and the tuner panel both persist past teardown. If you'd rather gate them behind a dev flag for production, easy follow-up.

PR #692, opened 2026-07-01
fix: correctly import and notate multi-staff (piano/keys) tracks from GP8

OmikronApex:
## Summary
- Fix GP8 bass stave being silently dropped on import due to bar-column enumeration bug in `_gpif_tracks` / `filtered_to_column`
- Fix `notation_lift.split_hands` using a largest-gap heuristic that misassigned notes to the wrong hand for chords straddling middle C

## Test plan
- [ ] Import a GP8 file with a two-stave piano/keys track
- [ ] Verify bass stave notes appear (not just treble)
- [ ] Verify LH/RH split is correct for chords crossing middle C


topkoa (commented):
## Review — multi-stave GP8 import (companion to got-feedback/feedBack-plugin-editor#38)

**The core fix is correct.** I traced the stave-column semantic change (`MasterBar/Bars` = one column per *stave*) across every consumer: all `bar_ids[...]` lookups in the main loop, the vocal/drum paths, and `gp2notation.convert_track_to_notation` (whose docstring specifies Bars-column semantics for `raw_idx`/`lh_raw_idx`) are consistent with the new mapping. The per-stave tuning parse fixes a real silent-note-drop, and the bar-column comments are accurate and useful.

Inline comments below cover: one confirmed behavioral regression in `split_hands` (with a suggested guard that keeps this PR's fix), a stave-coverage gap that breaks the `note_count` preview invariant, and some duplication that will drift. Cross-cutting items that don't anchor to a line:

1. **`lib/gp_autosync.py` still carries both bugs this PR fixes.** Its `_synthesise_score_chroma` uses a raw `enumerate(Tracks)` index as the Bars column (lines 218/270) and the last-Tuning-wins `.//Property` scan (line 242) — and its docstrings explicitly claim parity with `gp2rs_gpx` ("the same ordering gp2rs_gpx._note_midi and _gpx_tuning rely on"), parity this PR silently breaks. It has no production call site today (CLI + tests only), so a follow-up issue is fine, but the fix is mechanical since this PR already contains the exact pattern.
2. **DCO**: the commit has no `Signed-off-by` trailer — CONTRIBUTING.md requires it; please recommit with `git commit -s --amend`.
3. **CHANGELOG.md**: `[Unreleased]` isn't updated; this is a user-facing bugfix (GP8 bass stave silently dropped) so it should get an entry per the repo convention.

Recommend addressing the `split_hands` guard and the stave-coverage/`note_count` mismatch before merge; the rest can be follow-ups.

*Review process: 8 independent analysis passes over the diff + enclosing code, candidates independently re-verified against the PR head; companion review posted on got-feedback/feedBack-plugin-editor#38.*

lib/notation_lift.py
```diff
@@ -114,11 +114,18 @@ def split_hands(notes: list[dict]) -> dict[str, list[dict]]:
...
-            gaps = [pitches[i + 1] - pitches[i] for i in range(len(pitches) - 1)]
-            split_after = gaps.index(max(gaps))
-            threshold = pitches[split_after]  # lh: midi <= threshold
+            # Prefer middle C as the split boundary when notes straddle it —
+            # this correctly handles bass+treble chords from piano imports where
+            # the largest-gap heuristic picks the wrong split point (e.g.
+            # [G2, E3, C4]: largest gap is G2→E3 but the real split is E3|C4).
+            # Fall back to the largest internal gap only when all notes land on
+            # one side of middle C.
+            if pitches[0] < MIDDLE_C <= pitches[-1]:
```
topkoa, line 128: **Confirmed regression class: the hard middle-C split can emit a physically unplayable hand.**

For `[E2=40, B3=59, D4=62, G4=67]` — a bass note under an Em7-shape RH voicing that dips just below C4, an extremely common pop/piano figure — the old largest-gap split gave `lh=[E2]`, `rh=[B3,D4,G4]` (largest gap 40→59 = 19 semitones). The new rule gives `lh=[E2,B3]`: a simultaneous **19-semitone LH span**, exceeding the very 12-semitone `HAND_SPLIT_SPAN_SEMITONES` that triggered the split. The branch never re-checks the spans of the sub-groups it produces. SATB/hymn voicings mis-assign the same way (alto routinely sits below C4 but is engraved treble/RH: `[G2,D3,B3,G4]` → old correctly gave `rh=[B3,G4]`, new dumps B3 into LH).

Your `[G2,E3,C4]` example is a genuine fix, so don't revert — **guard it**: use the middle-C boundary only when both resulting groups are themselves playable, else keep the gap heuristic:

```python
threshold = None
if pitches[0] < MIDDLE_C <= pitches[-1]:
    lh = [p for p in pitches if p < MIDDLE_C]
    rh = [p for p in pitches if p >= MIDDLE_C]
    if (lh[-1] - lh[0] <= HAND_SPLIT_SPAN_SEMITONES
            and rh[-1] - rh[0] <= HAND_SPLIT_SPAN_SEMITONES):
        threshold = MIDDLE_C - 1
if threshold is None:
    gaps = [pitches[i + 1] - pitches[i] for i in range(len(pitches) - 1)]
    threshold = pitches[gaps.index(max(gaps))]
```

This keeps `[G2,E3,C4]` → `lh=[G2,E3]` (LH span 9, RH span 0) while restoring `[E2,B3,D4,G4]` → `lh=[E2]`. A test with exactly that Em7 chord would pin it — the existing test at `tests/test_notation_lift.py:71-76` uses `[36, 72]`, where both heuristics agree.

Worth noting the blast radius: after the companion PR, GP imports bypass `notation_lift` entirely, so this heuristic permanently governs **MIDI keys imports, hand-authored charts, and every re-save after a sloppak reopen** (see the lifecycle comments on the companion PR) — non-GP content this fix was never aimed at.

lib/gp2rs_gpx.py
```diff
@@ -1979,6 +2020,94 @@ def _auto_guitar_hint(track_idx: int):
             # auto-select (which keys on arr_name.startswith("keys")) still work.
             arr_name = re.sub(r'\s*\d+$', '', arr_name).strip() or 'Keys'
 
+        elif is_keys and track.get('num_staves', 1) > 1:
+            # GP8 two-stave piano: stave 1 (bass clef) is a second column in
+            # MasterBar/Bars for the same Track element.  Walk it exactly like
+            # the GPX LH merge above and fold its notes into the arrangement.
+            _s1_col = track['stave_columns'][1]
```
topkoa: **Staves beyond index 1 are silently dropped** — this fold hardcodes `stave_columns[1]`, and the notation-sidecar call below (line ~2170) likewise passes only `stave_columns[1]` as `lh_raw_idx`. A track with 3+ `<Staves>/<Staff>` entries (hand-crafted GPIF, third-party exporters, or a future GP organ/pedal layout) loses stave 2+ from both the arrangement and the notation payload, with no warning.

GP8's own UI likely caps grand staff at 2, so the 3-stave case is mostly defensive — but the fix is nearly free and makes the code self-consistent with `note_count` (which sums *all* columns, see the comment at line 375):

```python
for _s1_col, _s1_sp in zip(track['stave_columns'][1:],
                           track['stave_pitches'][1:]):
    ...  # existing walk body
```

The zip also removes the separate `len(stave_pitches) > 1` re-check — `_gpif_tracks` guarantees the two lists are parallel.

lib/gp2rs_gpx.py
```diff
@@ -319,27 +327,54 @@ def _note_count_for_raw(raw_idx: int) -> int:
...
-            'string_pitches': string_pitches,
+            'string_pitches': stave_pitches[0],   # primary stave (existing key)
+            'num_staves': num_staves,
+            'stave_columns': stave_columns,
+            'stave_pitches': stave_pitches,
             'is_drums': is_drums,
             'midi_program': midi_program,
             'midi_channel': midi_channel,
-            'note_count': _note_count_for_raw(raw_idx),
+            'note_count': sum(_note_count_for_raw(c) for c in stave_columns),
```
topkoa, line 374: **This breaks the invariant `_note_count_for_raw`'s own comment promises** ("excluding [tie destinations] makes the preview match what the user actually gets"): the count now sums *all* stave columns, but conversion only walks extra staves for `is_keys` tracks. A 2-stave grand-staff track whose instrument evades the keys heuristic (celesta, glockenspiel, harp, marimba, synth pads — GM programs outside `{0-7, 16-23, 80-83}` — with a non-keyword name) classifies as guitar, converts stave 0 only, and the import dialog advertises notes that then vanish.

Two consistent options:
- count only the columns conversion will actually walk (stave 0, plus staves 1+ when the track would take the keys path), or
- (better) fold staves 1+ for *all* multi-stave tracks, not just keys — the note walk doesn't care about instrument.

Separately: each `_note_count_for_raw(c)` call re-walks the entire MasterBars list and re-splits each `Bars` text. Counting all of a track's columns in a single masterbar pass would avoid N full sweeps per multi-stave track.

lib/gp2rs_gpx.py
```diff
@@ -1366,16 +1401,22 @@ def convert_file(
...
 
-    # Map filtered track index -> raw track index (needed for bar lookup)
+    # Map filtered track index -> bar column (MasterBar/Bars position for
+    # stave 0 of that track).  Each <Staves/Staff> child occupies one column,
+    # so multi-stave tracks advance the column counter by num_staves, not 1.
+    # The old enumerate()-based mapping assumed 1 stave per track and produced
+    # wrong columns for any track following a multi-stave predecessor.
     raw_tracks = list(root.find('Tracks') or [])
-    filtered_to_raw: dict[int, int] = {}
+    filtered_to_raw: dict[int, int] = {}   # kept as name for caller compat
```
topkoa: **The stave-counting rule now lives in two places in this file.** This loop re-derives `num_staves` / pseudo-track skip / bar-column arithmetic that `_gpif_tracks` already computed and stored on every track dict. Since `tracks` is in scope here, the whole loop is:

```python
filtered_to_raw = {i: t['stave_columns'][0] for i, t in enumerate(tracks)}
```

One source of truth matters here specifically because divergence in stave counting *is* the bug class this PR fixes — if the counting rule ever changes (a Staff variant that must be filtered, pseudo-track detection tweaks), the two copies drift and every track after a multi-stave one reads the wrong column again. Also saves a second full DOM scan of `Tracks`.

(While here: the name is now misleading — it maps to a bar *column*, not a raw track index. A rename to `filtered_to_col` with a one-line alias comment would prevent the next reader from indexing `raw_tracks` with it.)

lib/gp2rs_gpx.py
```diff
@@ -1979,6 +2020,94 @@ def _auto_guitar_hint(track_idx: int):
             # auto-select (which keys on arr_name.startswith("keys")) still work.
             arr_name = re.sub(r'\s*\d+$', '', arr_name).strip() or 'Keys'
 
+        elif is_keys and track.get('num_staves', 1) > 1:
```
topkoa: **This ~85-line block is a near-verbatim copy of the GPX LH-merge branch above** (~lines 1910-2015) — the dedup comment below even says "Same dedup-merge as the GPX LH path above." The only real differences are the column index and the tuning list.

That's timing-critical logic duplicated (tempo-map advance, time-sig fallback, tie sustain extension, the 0.2s sustain threshold, the `(round(time,3), string, fret)` dedup key): any future fix to one copy — a tie bug, repeat expansion, dedup-precision change — must be manually mirrored or GPX two-track and GP8 two-stave imports silently diverge, producing different timings for the same score depending on file format. That class of bug only manifests on one format and is miserable to trace.

Suggest extracting before this pattern gets a third copy:

```python
def _collect_column_notes(col, string_pitches) -> list[RsNote]:
    ...  # the shared walk

def _merge_lh_notes(rs_notes, rs_chords, lh_notes) -> None:
    ...  # the shared dedup-merge (skip entirely when lh_notes is empty)
```

Both branches then reduce to computing `(col, pitches)` + two calls. (`gp2notation._walk_track_beats` already proves the walk skeleton is factorable.)

lib/gp2rs_gpx.py
```diff
@@ -319,27 +327,54 @@ def _note_count_for_raw(raw_idx: int) -> int:
...
+                    if prop.get('name') == 'Tuning':
+                        pe = prop.find('Pitches')
+                        if pe is not None and pe.text:
+                            try:
+                                pitches = [int(p) for p in pe.text.split()]
+                            except ValueError:
+                                pass
+                        break
+                stave_pitches.append(pitches)
+        if not stave_pitches:
```
topkoa: Minor hardening: when `<Staves>` exists with Staff children but *no* staff carries a Tuning property, `stave_pitches == [[]]` — truthy — so this track-level fallback never runs and the track ends up with empty `string_pitches` (fretted notes drop silently in `_note_midi`). I checked the known producers (GP6 has no `<Staves>` at all; GP7/GP8 and alphaTab always write per-staff Tuning), so no current writer emits that shape — flagging it because the failure mode is silent note loss. `if not any(stave_pitches):` (or a per-staff fallback to the track-level property) would close it.

Also, the per-staff loop and this fallback inline the identical 10-line Tuning/Pitches parse — a tiny `_parse_tuning(el) -> list[int]` used as `[_parse_tuning(s) for s in staves_el]` / `[_parse_tuning(t)]` would collapse the section (and `gp_autosync.py:242` has a third copy that could reuse it).

byrongamatos:
Addressed @topkoa's review (pushed as a follow-up commit; both commits now DCO-signed):

- **`split_hands` unplayable-hand regression** — fixed. The middle-C boundary is now used only when *both* resulting hands are within `HAND_SPLIT_SPAN_SEMITONES`; otherwise it falls back to the largest-gap heuristic. `[E2,B3,D4,G4]` → `lh=[E2]` again, while `[G2,E3,C4]` still splits at middle C. Pinned by two tests in `tests/test_notation_lift.py`.
- **`note_count` invariant / stave-coverage** — fixed at the root: a multi-stave (grand-staff) track is now classified as keys end-to-end, so stave-0 and folded stave-1+ share one encoding and the `note_count` sum-all-columns matches what imports — including the celesta/harp/marimba cases the name/program heuristics miss. GPIF writes guitars as a single Staff, so ordinary fretted tracks aren't swept in.
- **Staves beyond index 1** — fixed: the fold now iterates `stave_columns[1:]` (all extra staves), not just stave 1.
- **85-line LH-merge duplication** — extracted `_collect_column_notes` + `_merge_lh_notes`; the GPX LH/RH pair merge and the GP8 grand-staff fold now share one implementation.
- **`filtered_to_raw` duplication** — rebuilt from the already-computed `stave_columns` (one source of truth) and dropped the dead `raw_tracks`/`num_raw_tracks`. Kept the name (with a comment that it's a bar *column*) to limit blast radius rather than renaming ~6 call sites.
- **Tuning `[[]]` silent-loss** — closed via a `_parse_tuning` helper with a per-staff fall-back to the track-level property (parity-safe: `stave_pitches` stays parallel to `stave_columns`).
- **DCO + CHANGELOG** — both done.

Deferred (noted in the CHANGELOG): `lib/gp_autosync.py` still carries the pre-fix bar-column + tuning logic — CLI/tests only, no production caller, mechanical follow-up.

PR #694, opened 2026-07-01
perf: eliminate per-frame layout thrash, shader-compile spikes, and per-frame allocations in the player

OmikronApex:
## Why

Users reported performance problems with the 3D highway. A Chrome performance trace (63 s session) showed 100 dropped frames (worst 501 ms) and identified the actual offenders — mostly **core**, not the renderer:

- `_isHighwayVisible` read `canvas.offsetParent` every rAF frame (forces style/layout recalc) — ~456 ms self-time.
- The v3 player-chrome rAF loop called `matches(':hover')` per frame and unconditionally rewrote the Up-Next pill's `textContent`/bar width at 6 Hz — ~1.5 s incl. ~1800 layout passes.
- Frame spikes correlated with Three.js first-use shader compilation and lazy label-texture uploads (`getParameters`, `texSubImage2D`).

## What

**Commit 1 — core layout thrash (trace-proven, benefits all viz):**
- `highway.js`: DOM visibility sampled every 10th frame with a cached value; fresh sample forced on init / canvas replace / resize / override-clear so hide-show transitions stay correct.
- `player-chrome.js`: hover tracked via mouseenter/mouseleave; Up-Next element refs cached, DOM written only on value change (eta coarsened to whole seconds beyond 10 s); progress bar moved from `width` to compositor-only `transform: scaleX` (+ `v3.css` transform-origin).

**Commit 2 — spike smoothing (highway_3d):**
- `ren.compile(scene, cam)` + deterministic label textures (fret numbers in every per-frame style/colour combo) pre-warmed at init, during the load spinner; chart-dependent chord/section label textures pre-warmed on first draw. Plugin version 3.30.2 → 3.31.0.

**Commit 3 — allocation/scan hardening for weaker hardware (static analysis):**
- `_makeBundle` mutates one persistent per-instance object instead of allocating a fresh ~35-field bundle per frame (×N under splitscreen). Object identity stable + meaningless; array fields still swap reference on chart changes (field-identity caches unaffected). Contract documented in both CLAUDE.mds.
- New `bsearchTime` windows the default 2D renderer's beat scan (was O(all beats)/frame); `bundle.lowerBoundT` / `bundle.lowerBoundTime` expose the searches so custom viz stop reimplementing culling.
- highway_3d: `localStorage` sustain flag polled at ~1 Hz instead of per frame; lyrics `measureText` layout cached per displayed line.
- `tests/js`: bundle source-shape assertions widened to accept the assignment form.

Out of scope (deliberately): InstancedMesh draw-call rework, chord-diagram cache shrink (tracked in-code follow-up).

## Measured results (same machine, re-traced)

| Metric | Before | After (song A) | After (dense 16ths song B) |
|---|---|---|---|
| Median rAF frame | 3.19 ms | 0.64 ms | 1.08 ms |
| p95 frame | 16.6 ms | 2.35 ms | 1.63 ms |
| Worst frame | 501 ms | 7.8 ms | 5.6 ms |
| Dropped frames (>16.7 ms) | 100 | 0 | 0 |

`_isHighwayVisible` and the chrome `loop` left the profile's top 40 entirely; shader-compile spikes gone; rAF now runs at full monitor refresh.

## Testing

- `pytest`: 1846 pass; 12 failures reproduce on a clean tree (pre-existing tuner/diagnostics/tailwind issues, unrelated — this diff is frontend-only).
- `npm run test:js`: 947 pass / 16 fail — identical to the clean-tree baseline.
- Manual: v2 + v3, default 2D + 3D highway, splitscreen hide/show, Up-Next pill, transport auto-hide, lyrics.


static/highway.js
```diff
@@ -733,103 +746,121 @@ function createHighway() {
...
+    // _makeBundle() mutates its fields in place each call instead of
+    // allocating a fresh ~35-field object per rAF frame (steady GC churn
+    // on weak hardware, ×N under splitscreen). Consequences for
+    // consumers: the bundle OBJECT's identity is stable across frames
+    // and carries no meaning; its field values are only valid for the
+    // duration of the current draw call. Array fields (`notes`,
+    // `chords`, `handShapes`, `chordTemplates`, ...) still swap
+    // reference whenever chart data changes — field-identity caches
+    // (e.g. highway_3d's merge caches) rely on that invariant.
+    const _bundleReused = {};
```
byrongamatos, line 764: **[1 — headline, needs external-repo audit] Reused bundle object breaks any external viz that retains it past the synchronous draw call.**

`_makeBundle()` now mutates this shared object; scalar fields (`currentTime`, `isPlaying`, `renderScale`, `getNoteState` verdicts) are valid only for the current draw call. In-repo `highway_3d` is safe (reads synchronously; caches only array-field references, which still swap on chart change). But feedBack ships external bundled viz — staffview, tabview, piano, drums, keys_highway_3d, drum_highway_3d — in separate repos. Any that stash the whole bundle (`this._b = bundle`) and read a scalar field a frame later, or in a `setTimeout`/`Promise`/`rAF`, now read a *different* frame's mutated values (time drift, wrong `isPlaying`). The contract is documented in two CLAUDE.mds but nothing enforces it. Please grep those repos for cross-frame bundle retention before merging.

plugins/highway_3d/screen.js
```diff
@@ -9886,6 +9908,55 @@
...
+        //                           names, section names); needs the ready
+        //                           bundle, so it runs once from the first
+        //                           draw() after each init.
+        // txtMat() rasterises into the unbounded cache these draws hit
+        // anyway; ren.initTexture() forces the GPU upload now.
+        let _chartPrewarmed = false;
+        function _prewarmTex(mat) {
+            if (mat && mat.map && ren) ren.initTexture(mat.map);
+        }
+        function _prewarmStatic() {
```
byrongamatos, line 9968: **[2 — confirmed] Pre-warm misses the `teachMark` and `technique` label styles (and color-variant `noteFret`/`chord`).**

The draw loop calls `txtMat` with styles `noteFret, fretRow, ghostFret, chord, section, teachMark, technique`, but `_prewarmStatic`/`_prewarmChart` only warm gold/idle `noteFret`/`fretRow`, white `ghostFret`, and fixed-color `chord`/`section`. Teaching-mark (`#7fd1ff`) and technique labels — plus verdict-tinted `noteFret` — still compile/upload lazily on first appearance, so those `texSubImage2D` spikes remain mid-song despite the stated goal of moving *all* first-appearance costs into the load spinner. It's also a maintenance trap: a new label style silently reintroduces a spike, with no test linking the two lists.

static/v3/player-chrome.js
```diff
@@ -230,8 +272,8 @@
         if (!p) return;
         const playBtn = $('btn-play');
         const playing = playBtn && playBtn.getAttribute('aria-pressed') === 'true';
-        const controls = $('player-controls');
-        const overControls = controls && typeof controls.matches === 'function' && controls.matches(':hover');
+        // overControls maintained by mouseenter/mouseleave (see start()) —
```
byrongamatos, line 275: **[3 — plausible] Cached `overControls` hover boolean can get stuck true or go permanently stale.**

Edge-driven `mouseenter`/`mouseleave` replaces per-frame `matches(':hover')`. If `#player-controls` is hidden/detached while the pointer is over it (missed `mouseleave`), `overControls` stays `true` and `tickIdle` never lets the transport auto-hide during playback. Conversely, listeners bind once to the node resolved in `start()`; if `#player-controls` is re-created without a `stop()`/`start()` cycle, hover tracking goes dead. The old `matches(':hover')` self-corrected every frame. Cheap fix: reconcile `overControls` from `matches(':hover')` on the ~1 Hz Up-Next tick, or re-resolve the node the way `_upnextRefs()` does via `isConnected`.

static/highway.js
```diff
@@ -1246,7 +1278,14 @@ function createHighway() {
     // hosts that need those use setVisible() instead.
     function _isHighwayVisible() {
         if (_visibleOverride !== null) return _visibleOverride;
-        return !!(canvas && canvas.offsetParent !== null);
+        // Throttled offsetParent read — see _DOM_VIS_CHECK_FRAMES above.
+        if (Number.isNaN(_domVisSampledFrame)
+            || ((_frameIdx - _domVisSampledFrame) | 0) >= _DOM_VIS_CHECK_FRAMES
+            || ((_frameIdx - _domVisSampledFrame) | 0) < 0) {
+            _domVisCached = !!(canvas && canvas.offsetParent !== null);
```
byrongamatos, line 1290: **[4 — confirmed, minor] Public `highway.isVisible()` now returns this throttled cache, contradicting its own docstring.**

`isVisible()` (~line 3847) is documented as "the live DOM check … late subscribers can call this once to sync their initial state," but it returns `_isHighwayVisible()`, sampled at most every 10 frames. An external renderer that binds to `highway:visibility` after a transition and calls `isVisible()` to seed state can get a value up to ~166 ms stale until the next throttled emit self-heals it. Either force a fresh sample inside `isVisible()` (set `_domVisSampledFrame = NaN` before the read) or fix the docstring.

static/highway.js
```diff
@@ -733,103 +746,121 @@ function createHighway() {
...
+        // closure ctx, which is null for custom renderers. Renderers
+        // that need lefty-aware text should check `bundle.lefty` and
+        // apply the mirror transform themselves on their own context.
+        b.project = project;
+        b.fretX = fretX;
+        // Windowed-iteration helpers (stable references): lower-bound
+        // binary searches so custom viz don't full-scan chart arrays per
+        // frame. lowerBoundT keys on `.t` (notes / chords); lowerBoundTime
+        // keys on `.time` (beats / anchors / sections).
+        b.lowerBoundT = bsearch;
```
byrongamatos, line 851: **[5 — cleanup] New `bundle.lowerBoundT`/`lowerBoundTime` have no in-tree consumer while highway_3d keeps its own byte-identical local `lowerBoundT`.**

The helpers are added so viz "stop reimplementing culling," but the one in-repo viz still uses its local duplicate (`plugins/highway_3d/screen.js:~1239`) at ~30 call sites and never reads these fields. Either adopt `bundle.lowerBoundT` in highway_3d and delete the local helper, or note the fields are speculative API. Relatedly, `bsearchTime` (line 2676) is a third near-identical copy of `bsearch`/`bsearchChords` differing only by key field — a single `lowerBound(arr, val, key)` would collapse all three, and the new fields do no validation (silently return `0` if handed the wrong-keyed array).

static/highway.js
```diff
@@ -122,6 +122,19 @@ function createHighway() {
...
+    // every rAF frame forces a style/layout recalc — profiled at ~0.5 s
+    // main-thread self-time over a 63 s session. The displayed state
+    // changes rarely (navigate / splitscreen panel toggle), so the DOM
+    // is only re-sampled every _DOM_VIS_CHECK_FRAMES frames; the cached
+    // value serves the frames in between (worst-case transition latency
+    // ~10 frames ≈ 166 ms at 60 Hz — fine for a hide/show pause signal).
+    // Set _domVisSampledFrame to NaN to force a fresh sample on the next
+    // check (done on init, canvas replace, resize, and override-clear so
+    // deliberate transitions don't wait out the throttle window).
+    const _DOM_VIS_CHECK_FRAMES = 10;
```
byrongamatos, line 140: **[6 — altitude/maintainability] The DOM-visibility throttle depends on manual `_domVisSampledFrame = NaN` invalidation at four sprinkled sites** (init, canvas-replace, resize, override-clear).

Any future visibility-affecting path — a new embed, tab-restore, programmatic re-parent, or a splitscreen `display:none` toggle that doesn't call `resize()` — that forgets the reset silently serves stale visibility for up to ~166 ms, with no test to catch the omission. An `IntersectionObserver` would fire on transitions natively and remove both the per-frame `offsetParent` read and the manual invalidation entirely.

byrongamatos (commented):
Review of the perf pass (manual multi-angle + independent Codex review). The measured wins are real and the diff is clean overall — no crash-class bug. Findings are posted inline, ranked by severity. The one gating item is the reused-bundle contract (comment on `static/highway.js:759`): safe in-repo, but it needs a quick audit of the external bundled-viz repos before merge.

_Verified NOT bugs during review:_ `_chartPrewarmed` **is** reset per song (`playSong` → `highway.init()` → `_setRenderer` → `renderer.init()` runs unconditionally, so `_prewarmChart` does run for song 2+); `drawBeats` windowing fully covers `project()`'s drawable range; splitscreen uses separate `createHighway()` instances so the reused bundle can't clash across panels; the `_frameIdx` wrap guard is only reachable after ~414 days.

OmikronApex:
Review findings addressed in `1e0…` (review fixes) + the follow-up perf commit:

**[1] External bundle-retention audit — clean.** Shallow-cloned and grepped all six external bundled-viz repos (staffview, tabview, piano, drums, keys_highway_3d, drum_highway_3d) for whole-bundle retention (`x = bundle`, `this.bundle =`) and bundle-object identity checks (`bundle !==`): zero hits. All bundle access is synchronous per-draw field reads; the async code in those repos is MIDI/script loading, unrelated to the bundle. No external repo changes needed.

**[2] Fixed.** Pre-warm now covers teachFg/teachSd label textures and the technique sprite factories (PM/FH mute X, hammer/pull triangles, bend chevron stacks, slide arrows) per active-palette string colour, with a maintenance note tying new label styles to the warm list. (Verdict-tinted colours remain dynamic by nature — but see the follow-up commit: map swaps no longer recompile at all, which removes most of the residual spike risk.)

**[3] Fixed.** `overControls` is reconciled against `matches(':hover')` on the existing ~6 Hz throttled tick — unsticks a missed mouseleave and survives a re-created `#player-controls` node.

**[4] Fixed.** `isVisible()` forces a fresh DOM sample; the docstring's 'live DOM check' contract holds again.

**[5] Documented.** highway_3d keeps its local `lowerBoundT` deliberately (downlevel hosts + ~30 call sites incl. bundle-less helpers); a comment at the local helper and the plugin CLAUDE.md point new code at `bundle.lowerBoundT`/`lowerBoundTime`.

**[6] Documented rather than rearchitected.** The manual `NaN` invalidations are latency optimizations only — the periodic 10-frame resample is the ground truth, so a forgotten reset self-heals within ~166 ms and can never serve stale visibility indefinitely (comment added at the constant). IntersectionObserver would also change semantics (viewport intersection vs display:none), so deferred.

**Follow-up commit** (from a 4×-CPU-throttled retrace with stack attribution): label map swaps no longer set `material.needsUpdate` (program unchanged between two non-null textures — was ~4% of throttled main thread in `getParameters`), the 3D highway's per-frame `getBoundingClientRect` self-check is throttled to every 10th frame (backing-store changes still re-apply same-frame, with a shape test asserting that), and the core 60 Hz HUD clock writes `textContent` only on change. Remaining known writer is notedetect's `badges.js` (~220 ms) — external repo, will file there.

Test suites at clean-tree baseline: pytest 1846 pass / 12 pre-existing failures, test:js 947 pass / 16 pre-existing.

PR #722, opened 2026-07-02
library: scraper options — per-source + per-field auto-apply, review-queue order (R1)

ChrisBeWithYou:
## What

The full scraper-options panel (R1 of the round-2 library-metadata batch), grown out of the Settings → Library "Metadata matching" card that shipped with #710. Not everybody needs the same things out of a scraper — this makes each piece opt-out:

**Sources — who may be contacted**
- `enrich_src_musicbrainz` — gates the background matcher (phase 2). Identity hashes still stamp each pass, and manual Fix-match / "Search instead…" in the review modal keep working, exactly like the existing master toggle's contract.
- `enrich_src_caa` — gates the Cover Art Archive fetch (phase 3). Rows skipped while a toggle is off stay unevaluated (`art_state` NULL / `unscanned`), so re-enabling picks them up on the next pass — nothing is permanently forfeited.

**Auto-apply — what a confident match may fill in on its own**
- `enrich_apply_names` / `enrich_apply_year` / `enrich_apply_genres` filter what an **automatic** match canonicalizes, on all three automatic paths (content-hash cache copy, manifest mbid/isrc exact keys, text auto tier). MusicBrainz ids always stamp — they're identity, not display; the art fetch and future re-matching need them.
- A match the user **confirms in the review modal applies in full** — the toggles gate only what happens without them.
- `enrich_apply_art` gates the art fetch together with the CAA source toggle. Two axes for one behaviour today, deliberately: a future second art source slots in without re-teaching the panel.

**Review queue**
- `enrich_review_order` = `missing_first` (default — today's behaviour) | `artist` | `recent`, consumed by `GET /api/enrichment/review`. Unknown stored values degrade to the default.

**UI** — the card gains Sources / Auto-apply / Review-queue-order groups (wired in `match-review.js`); the master toggle is relabelled "Match songs automatically" so it doesn't read the same as the new MusicBrainz source checkbox. No tailwind rebuild needed — every utility used was already in the scanned set.

## Stacking

Based on `feat/enrichment-art` (#715) — the art toggles gate that PR's CAA worker. **Merge #715 first**, then this lands clean.

## Verification

- `tests/test_scraper_options.py` (9 new): settings validation; MB-source-off stamps hashes but never calls the (fake) transport, and re-enabling matches the same row; per-field stripping on auto matches with ids preserved; review-accept applies all fields despite toggles; CAA gating on both axes; all three review-order modes plus the unknown-value fallback.
- All enrichment suites green (77 tests across P7/P8/R3/R1 + pure matcher).
- Full-suite failure set is **byte-identical** with the change stashed vs applied (39 known env/pre-existing failures) — A/B proven.
- `node --check` clean; no new tailwind classes (verified against the built stylesheet).

https://claude.ai/code/session_01Nm7tHs1Yvjjtnnu4nzJgdN

### DIFF
--- a/server.py
+++ b/server.py
@@ -874,9 +874,18 @@ class MetadataDB:
         # MusicBrainz just to render; `last_attempt_at` anchors the failed-row
         # retry backoff (epoch seconds). Idempotent ALTERs, same pattern as
         # the `songs` migrations above.
+        # R1 scraper options: `apply_mask` records which per-field auto-apply
+        # toggles were OFF (suppressed) when an AUTOMATIC match settled the row,
+        # as a canonical sorted comma-joined marker of blocked keys (''/NULL =
+        # nothing suppressed). It keeps the per-field toggles to the same
+        # "nothing forfeited" contract as the source/art toggles: re-enabling a
+        # field re-queues affected `matched` rows for backfill (enrichment_pending)
+        # and a partially-applied row is barred from seeding siblings
+        # (enrichment_cache_lookup). Idempotent ALTER, same pattern as above.
         for ddl in (
             "ALTER TABLE song_enrichment ADD COLUMN candidates TEXT",
             "ALTER TABLE song_enrichment ADD COLUMN last_attempt_at REAL",
+            "ALTER TABLE song_enrichment ADD COLUMN apply_mask TEXT",
         ):
             try:
                 self.conn.execute(ddl)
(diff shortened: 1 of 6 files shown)


byrongamatos:
Superseded by #726 (merged). GitHub auto-closed this PR when its stacked base branch `feat/enrichment-art` was deleted on #715's squash-merge, and refused to reopen it. #726 is the identical branch rebased onto `main` — same content, including the reviewed P2 fix.

PR #733, opened 2026-07-03
v3 library: per-field pack overwrite (R4b) — fix wrong author values, with receipts [DRAFT: gated on spec §7]

ChrisBeWithYou:
> **DRAFT — do not merge yet.** Gated on the feedpak-spec §7 amendment (got-feedback/feedpak-spec#47). §7 today permits adding absent keys only (gap-fill, shipped as #724). Replacing an author-set value needs topkoa's sign-off on the amendment. This PR is built **fully compliant with the proposed amendment shape** so it can land the moment §7 is amended.

Gap-fill (#724) fills *absent* keys; this fixes **wrong** author-set values (a typo'd artist, a wrong year) **one field at a time, with provenance and an undo**.

### Server (extends the merged gap-fill endpoints, doesn't fork)
- Preview `GET /api/song/{fn}/gap-fill` now also returns **`differs`** = `[{key, current, proposed}]` over the allowlist (title/artist/album/year/genres) where the manifest has an author value and the confirmed match supplies a *different* one — plus `overwrite_allowed` and `has_backup`. **`differs` is empty unless `match_state == 'manual'`**: a user-confirmed pin can authorize a replace; an auto-match never can. **`mbid`/`isrc` are never in `differs`** (identity changes only via re-match).
- `POST` accepts `overwrite_keys` alongside gap-fill's `keys`; gated on the `allow_pack_overwrite` setting + manual state + the allowlist, proposals **recomputed under the IO lock** so a field that changed since preview is skipped. Every written key (gap-fill **and** overwrite) lands in a new **`write_log`** receipts table, pruned to the newest 5000 rows.
- `POST /api/song/{fn}/revert-original` restores the pack from its `.bak` (dir + zip), re-syncs the DB; 404 when no backup; the `.bak` is **preserved** after revert. Demo-blocked.
- `GET /api/song/{fn}/write-log` — the song's receipts.
- Setting `allow_pack_overwrite` (**default OFF**).

### The `.bak` guarantee
Unchanged from the shipped writer: the backup is written **once** (first write) and is always the **pristine author original** — a second overwrite never clobbers it, so Revert always returns the untouched pack.

### Frontend
Details drawer *"Overwrite existing fields"* sub-block: per-key rows `Artist: "ACDC" → "AC/DC"`, **all unticked by default**, warning that it replaces the author's value and keeps a backup. When `differs` exist but the setting is off, one muted line points to Settings. A *"Revert file to original…"* link appears whenever a backup exists. Settings gains the `allow_pack_overwrite` checkbox.

### Verification
`tests/test_pack_overwrite.py` — differs manual-only + identity-excluded + equal-skipped; refused when setting-off / not-manual / bad-keys; dir + zip overwrite; **`.bak` stays the pristine original across a second write**; write_log records old+new (and gap-fills); prune at cap; revert dir+zip+404; demo blocks write & revert. **30 pass** with `test_gap_fill` green. `node --check` clean; no new Tailwind classes.

### Merge sequence
1. topkoa amends §7 on feedpak-spec#47 →
2. mark this ready for review → merge.

https://claude.ai/code/session_01Nm7tHs1Yvjjtnnu4nzJgdN

### DIFF
--- a/lib/songmeta.py
+++ b/lib/songmeta.py
@@ -49,6 +49,15 @@ def _apply_to_sloppak_manifest(manifest: dict, fields: dict) -> bool:
     if "year" in fields:
         manifest["year"] = _coerce_year(fields["year"])
         dirty = True
+    # `genres` is the feedpak list field (spec 1.12.0). Accepts a list/tuple
+    # (stringified item-wise) or a single name (wrapped); None leaves the
+    # existing value alone, mirroring the string fields above. Sent by the
+    # overwrite lane (R4b) — the manual Edit Metadata path never includes it.
+    if fields.get("genres") is not None:
+        raw = fields["genres"]
+        manifest["genres"] = ([str(g) for g in raw]
+                              if isinstance(raw, (list, tuple)) else [str(raw)])
+        dirty = True
     # Opportunistically declare the format version (spec §4) when we're already
     # rewriting because a metadata field was supplied. Gated on `dirty` (i.e. a
     # field was given) so this never forces a *standalone* rewrite with no fields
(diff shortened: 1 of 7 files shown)


ChrisBeWithYou:
**Integration note** — from a full-queue dry-run that merged every open org PR into a next-version test build.

Merges with two small adjustments:

1. **Rebase on `main`** — `main` added an `optInToggles` mechanism for default-OFF toggles. Move `allow-pack-overwrite` into `optInToggles` instead of the current inline special-case in `server.py` (functionally identical, and it clears the conflict).
2. **`static/v3/index.html`** — the settings rows for "Artist pages" (from `main`) and "Writing to your files" (this PR) both need to survive; rebalance the surrounding `div` nesting so both render.

Otherwise clean: the `write_log` table + `allow_pack_overwrite` config default + the validation-tuple key all combine with `main`'s artist-pages/AcoustID work without issue.

PR #753, opened 2026-07-04
settings: add host instrument profiles

ChrisBeWithYou:
## Summary
- add host-owned guitar and bass instrument profiles with lead, rhythm, and bass slots
- expand built-in tuning presets, including 6-string bass and lower guitar/bass tunings
- normalize settings import/export and keep legacy flat instrument settings mirrored
- let the instrument selector expose 6-string bass and fix working-tuning named preset seeding

## Tests
- node tests/js/working_tuning.test.js
- python AST syntax checks for touched Python files

## Notes
- python -m pytest tests/test_tunings.py tests/test_settings_api.py -q -p no:cacheprovider is blocked locally by missing dependency: structlog

## Summary by CodeRabbit

* **New Features**
  * Added instrument profile support to settings, including per-instrument tuning presets and profile-aware defaults.
  * Introduced an **Instrument pathway** dropdown (songs/practice/learn/studio) with persistence, hydration, and live reload.
* **Bug Fixes**
  * Improved startup tuning hydration to handle alternate `/api/tunings` payload shapes.
  * Expanded bass validation to support 6-string configurations.
* **Tests**
  * Updated and expanded coverage for profile/pathway defaults, migration, reset/export-import, and tuning conversions.

### DIFF
--- a/lib/tunings.py
+++ b/lib/tunings.py
@@ -207,7 +207,19 @@ def _valid_reference_pitch(value) -> float | None:
 
 def _valid_tuning_for_key(key: str, tuning):
     if isinstance(tuning, str):
-        return tuning if tuning in TUNING_PRESET_MIDIS.get(key, {}) and len(tuning) <= 64 else None
+        if len(tuning) > 64:
+            return None
+        if tuning in TUNING_PRESET_MIDIS.get(key, {}):
+            return tuning
+        # A name that IS a built-in preset for a different key is a misapplied
+        # built-in (e.g. "Drop D" on a 5-string bass, whose low string is B) —
+        # reject it. A name unknown to every built-in table is a provider/custom
+        # tuning (the tuner plugin's, exposed via /api/tunings) that this pure
+        # layer can't resolve — accept it so settings round-trip; the provider
+        # owns its validity.
+        if any(tuning in names for names in TUNING_PRESET_MIDIS.values()):
+            return None
+        return tuning
     if isinstance(tuning, list):
         expected = len(STANDARD_OPEN_MIDIS.get(key, []))
         if len(tuning) != expected:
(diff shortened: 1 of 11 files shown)


byrongamatos:
Review + fix pass (Claude Code). Deep-reviewed the instrument-profiles rework; it was **finding-dense**, so I fixed the CI-blocking + clear-correctness regressions and am flagging the rest. All settings/tunings suites green (138).

**Pushed 5 fixes:**
1. **Empty/unrelated POST froze default profiles into config.json** (broke `test_empty_post_preserves_all_existing_keys`). `save_settings` canonicalized on every save; now gated on the save actually touching instrument settings (GET already virtualizes profiles).
2. **Gameplay reset didn't reset `pathway`** — it's profile-mirrored, so deleting the flat key was a no-op (GET re-derived it). `reset_settings` now resets pathway inside the persisted profiles.
3. **Provider/custom tunings rejected** — the new per-profile validation only accepted built-ins, so tuner-plugin tunings (`/api/tunings`) errored on save and normalized to Standard on GET. `_valid_tuning_for_key` now accepts a name unknown to every built-in table while still rejecting a built-in misapplied to the wrong key.
4. **Fresh-config active-profile switch lost** — first-migration overwrote an explicit `active_instrument_profile` with the legacy-inferred one. Use `setdefault`.
5. **`test_instrument_fields_persist`** posted bass-5 + 'Drop D' (a 4-string tuning; 5-string bass drops to 'Drop A'). Updated the test.

**Two findings left for you — both are the same partial-update-semantics class and want a dedicated pass:**
- **[data-loss] Partial `instrument_profiles` update clobbers the others** (`server.py` ~9337): `normalize_instrument_profiles` fills omitted profiles with **defaults** and replaces wholesale, so `POST {"instrument_profiles":{"bass":{…}}}` resets a customized `guitar-lead`/`guitar-rhythm` to defaults. Needs merging the partial over the persisted profiles — which means deferring that normalization into the lock where the current cfg is available (a validation-flow restructure).
- **[client desync] `string_count`-only save silently rewrites tuning to Standard** (`tunings.py` ~383) when the old tuning is invalid for the new count, but the response is just "Settings saved" — the v3 string-count picker sends exactly this partial patch, so the browser/tuner keep the old tuning. Surface the effective tuning in the response (or have the client re-fetch).

Given the density here, I'd recommend a focused partial-update review of the profile layer before merge.

byrongamatos:
Follow-up: fixed the two partial-update issues I flagged earlier (139 settings/tunings tests green).

1. **Partial `instrument_profiles` no longer clobbers the others.** `save_settings` used `normalize_instrument_profiles(raw)`, which fills every omitted profile with defaults and replaces wholesale — so `POST {"instrument_profiles":{"bass":{…}}}` reset guitar-lead/rhythm. Now each *provided* profile is validated individually and the partial is merged over the persisted set inside the lock (a fresh config falls back to defaults for the unspecified ones). Test: `test_partial_instrument_profiles_update_preserves_others`.
2. **String-count switch no longer desyncs the tuning.** The string-count picker posted only `string_count`, so the backend silently reset a now-invalid tuning to Standard while the UI kept the old value. It now clamps + posts the valid tuning alongside `string_count`, mirroring the instrument-switch path (`static/v3/badges.js`).

PR #758, opened 2026-07-04
fix(enrichment): rank the canonical studio take over live/comp versions

ChrisBeWithYou:
## Problem
A flat MusicBrainz `/recording` text search ties every take of a song at the same score. Searching **AC/DC — "Highway to Hell"** returns ~10 live bootlegs and compilations with the canonical **1979 studio** version buried below the fetch limit. This hits the editor's new "Match metadata" popup *and* the background library enrichment (both share `lib/mb_match.py` + `/api/enrichment/search`).

## Fix (verified against live MusicBrainz)
- **`build_recording_query`** — drop live-**only** recordings (`-secondarytype:Live`). Compilations are deliberately kept: they *reuse* the studio recording, so filtering them cut the very recording we want (empirically confirmed).
- **`_best_release` / `parse_recording_doc`** — pick the canonical **studio album** (primary `Album`, no `Live`/`Compilation`/`Remix`/… secondary type) for the displayed album/year, and expose a `studio` flag.
- **`rank_candidates`** — the combined score caps at 1.0, so perfect-text-match candidates tie; break ties on the **studio** flag and, when the caller knows the audio length, on **duration proximity**. The studio distinction is *not* scored (a live take is still the right *song*) — only re-ordered.
- **`/api/enrichment/search`** — accept an optional `duration` param so a caller that has the audio but no library row (the editor's create modal, which holds the master track) can pass the length for the duration tiebreak.

**Result:** AC/DC "Highway to Hell" now returns the **1979 studio recording at #1** with the correct album + year.

## Tests
`tests/test_mb_match.py` + `tests/test_mb_enrichment.py` — 48 passing (query-string assertion updated for the new `-secondarytype:Live`).

## Summary by CodeRabbit

* **New Features**
  * Release matching now treats “clean studio albums” as canonical, prioritizing official studio releases over live or mixed types.
  * MusicBrainz enrichment search can use an optional track duration to strengthen matches when duration metadata is missing locally.
* **Bug Fixes**
  * Live recordings are filtered more reliably, while “(Live …)”/bracketed live chart titles still allow live takes through.
  * Candidate tie-breaking now accounts for whether the chart is live before applying duration-proximity selection.
* **Chores**
  * Increased the number of MusicBrainz recording candidates considered to improve downstream ranking quality.

### DIFF
--- a/lib/mb_match.py
+++ b/lib/mb_match.py
@@ -224,6 +224,11 @@ def _lucene_escape_phrase(s: str) -> str:
     return s.replace("\\", "\\\\").replace('"', '\\"')
 
 
+# A parenthetical/bracketed "(Live …)" marker — the live signal denoise() strips
+# from the title. Mirrors _NOISE_GROUP_RE but for the `live` term only.
+_LIVE_GROUP_RE = re.compile(r"[(\[][^)\]]*\blive\b[^)\]]*[)\]]", re.IGNORECASE)
+
+
 def build_recording_query(artist, title) -> str:
     """Lucene query for /ws/2/recording. Built from the DENOISED fields —
     the noise we strip (author credits, "(Live)", "(v2)") would otherwise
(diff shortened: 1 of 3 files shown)


byrongamatos:
Review + fix pass (Claude Code). The ranking logic is sound; pushed 3 fixes closing the live-vs-studio loop (26 mb_match tests green). CI red is the unrelated main-level starter-content bug.

1. **Official releases must outrank unofficial studio albums.** `_best_release` sorted `(clean, status_ok, date)`, so a bootleg Album beat an official Single/EP/comp — regressing canonical album/year and seeding cover-art from a bootleg for single-only songs. Reordered to `(status_ok, clean, date)`: official first, then prefer a clean studio album among official releases (still surfaces the studio album over an official live/comp album).
2. **Live charts kept their live recordings.** `build_recording_query` unconditionally added `-secondarytype:Live`, but `denoise()` strips a '(Live at …)' qualifier — so a genuinely-live chart had its only correct recording filtered out. Now skips the live filter when the source title has a parenthetical live marker (a bare word like 'Live and Let Die' still filters).
3. **Live charts no longer forced to the studio take.** Consequence of (2): `rank_candidates` still preferred the studio flag, so a live chart would auto-match the studio recording. Now drops the studio tiebreak for live charts — duration/score pick the right live version.

PR #759, opened 2026-07-04
feat(enrichment): AcoustID audio-fingerprint identification (opt-in)

ChrisBeWithYou:
## Why
Text search can only *guess* the version (see #758 for the ranking mitigation). The definitive fix is **content-based**: fingerprint the actual audio with Chromaprint (`fpcalc`) and look it up on **AcoustID**, which maps the fingerprint to the **exact** MusicBrainz recording — the approach Lidarr uses. This eliminates the studio-vs-live ambiguity entirely.

## What
- **`lib/acoustid_match.py`** — pure response parsing + config gating (unit-tested). Normalizes AcoustID hits into the same candidate shape as `mb_match`, so the review UI and the editor's Match popup render fingerprint and text hits identically. Flags the studio take (reuses the secondary-type logic).
- **`server.py`** — `_fpcalc` (Chromaprint subprocess), `_acoustid_lookup` (throttled, offline-guarded HTTP), `_identify_by_fingerprint` (also callable from the library-enrichment pipeline), and **`POST /api/enrichment/identify`** (upload the master audio → candidates in the `/search` shape).

## Opt-in + graceful
Absent the `fpcalc` binary **or** an `ACOUSTID_API_KEY`, the whole path is a no-op / `503` and the text matcher runs unchanged. Nothing regresses.

## Setup to enable
- `fpcalc` (Chromaprint) on `PATH` or at `$FPCALC`.
- A free AcoustID application key (https://acoustid.org/new-application) in `$ACOUSTID_API_KEY`.

## Tests
`tests/test_acoustid_match.py` — 8 passing (parsing, studio detection, dedupe, ranking, config gating). The `fpcalc` + live-lookup path needs the two deps above to exercise end-to-end.

## Follow-up
Wire an "Identify from my audio" button into the editor's Match popup (posts the staged master track to `/identify`); trivial once this endpoint lands + deps are configured.

## Summary by CodeRabbit

* **New Features**
  * Added “Identify by audio” to match review for both uploaded audio and library songs, returning enriched candidates (title, artist, album, year, duration, confidence).
  * Added optional AcoustID audio fingerprinting to Settings → Metadata matching (enable toggle + API key).
* **Bug Fixes**
  * Improved candidate ranking and de-duplication, with more accurate studio album selection and clearer handling when identification isn’t available or setup is incomplete.
* **Tests**
  * Added unit tests covering AcoustID response parsing and configuration gating.


byrongamatos:
Review + fix pass (Claude Code). Backend is solid — `_fpcalc` uses list-form subprocess args (no injection), `_song_audio_file` mirrors the sloppak containment guards, opt-in gating (412/503) is thorough, both endpoints are demo-blocked, 10 acoustid tests green. Pushed 2 fixes:

1. **`ci/tailwind-fresh` was red** — committed `static/tailwind.min.css` was stale vs a fresh rebuild. Regenerated with the pinned `tailwindcss@3.4.19` (verified byte-stable across rebuilds), so the CI check now passes.
2. **Identify upload read the whole body into memory** (`file.file.read()`). Now streams to the temp file in 1 MB chunks with a 256 MB cap (413 over) — fpcalc reads from the temp file anyway.

**Two follow-ups left for you:**
- **Pre-parse upload guard**: my cap runs after FastAPI has already spooled the multipart body. For full defense-in-depth, mirror `upload_song`'s pre-parse `Content-Length` check + `request.form(max_part_size=…)` (requires making the route async and running the blocking fpcalc/HTTP via `run_in_executor`). Low priority under the single-trusted-user model (Principle I), but it's the complete fix.
- **Settings UI to enable AcoustID**: the backend/API is complete (`save_settings` validates `acoustid_enabled`/`acoustid_api_key`, both in `_default_settings`), but the v3 settings card exposes no toggle/key input, so the 'Identify by audio' button 412s for a normal user until they POST settings or hand-edit config. Worth a small settings-card addition (a follow-up, since it'd also need a tailwind rebuild if it adds classes).

byrongamatos:
Follow-up: implemented both items I flagged (10 acoustid tests green; tailwind stays fresh).

1. **Pre-parse upload guard.** `/api/enrichment/identify` is now async: a pre-parse `Content-Length` check + `request.form(max_part_size=…)` reject an oversized body **before** Starlette spools the multipart to temp disk (mirrors `upload_song`), and the blocking fpcalc subprocess + AcoustID HTTP now run off the event loop via `run_in_executor` (the route was previously sync-in-threadpool; async keeps that non-blocking). Gate still fires before any parse (verified: unconfigured→412, configured-but-no-fpcalc→503).
2. **Settings UI to enable AcoustID.** Added an 'Identify by audio' opt-in toggle (`acoustid_enabled`, default OFF) + an AcoustID key input (`acoustid_api_key`) to the v3 Metadata-matching card, wired in `match-review.js` (loads current values, saves on change) — so the feature is reachable from the UI, not just a manual settings POST. Reused only existing classes, so `static/tailwind.min.css` stays fresh (ci/tailwind-fresh green).

PR #764, opened 2026-07-04
v3 library: "Refresh Metadata" button + per-view re-match + filename-artist seed

ChrisBeWithYou:
## What

A media-server-style **Refresh Metadata** control on the Songs toolbar (beside **⟳ Refresh**) — the metadata counterpart to a file scan. Where *Refresh* scans the folder for new files, this re-matches the songs you're **viewing** against MusicBrainz.

### Behaviour
- **Re-matches the visible grid window** — a per-view refresh, so it's visibly useful even on a library that's already been matched (the incremental background pass is a no-op there). The button **doubles as Stop** while a pass runs, with a **batch progress bar** and **per-tile `queued → working → done` badges**. **User-pinned (`manual`) matches are never touched.**
- **Filename-artist seed:** a lot of community charts ship with a **blank `artist` field**, which text search can't match on (and the per-field floor rejects). When the pack's artist is blank, the matcher now derives artist/title from the CDLC `Artist_Song-Title` **filename** convention as a **search seed only** — the displayed values still come solely from the confirmed MusicBrainz match, so nothing estimated is presented as author-set. This rescues blank-artist packs that previously always failed.

### API
- `POST /api/enrichment/rematch {filenames}` — reset the given (visible) songs to `unscanned` (skips `manual` via the existing never-overwrite guard) + kick one pass; returns the queued filenames.
- `POST /api/enrichment/cancel` — cooperative Stop; a cancel `Event` is checked between songs in the match + art phases.
- `POST /api/enrichment/states {filenames}` — per-window `{filename: match_state}` (+ current + running) for the tile badges. Read-only.
- `GET /api/enrichment/status` gains `total` / `matched` / `current` / `cancelling`.
- Demo mode: `cancel` + `rematch` blocked (mutate/spend rate budget); `states` stays open (read-only).

## Tests
`enrichment_states_for`, the three new routes, cancel-halts-a-pass-between-songs, kick-clears-a-stale-cancel, the filename parser, blank-artist seeding, and present-artist-not-overridden. Full enrichment + demo + scraper + context-menu + art suites green (162 passed).

## Verification
Exercised end-to-end on a live preview against a **1002-song DB + real MusicBrainz**: the pass progresses (running/total/matched/current at the ≤1/s throttle), **Stop halts mid-pass** (1 of 8 before breaking), cancel-when-idle is a clean no-op, and the filename seed rescued a real blank-artist pack (`failed → matched`). Frontend parses + is served correctly; the in-app visual pass (button/bar/tiles/toast pixels) is still worth an eyeball.

> Honest scope note: text enrichment has a real **data-coverage ceiling** — e.g. a library heavy in obscure romanized-Japanese city-pop returns 0 MusicBrainz candidates regardless of a correct artist/title seed. Those need audio-fingerprint (#759) or manual Fix-match (the metadata popup, next).

## Summary by CodeRabbit

* **New Features**
  * Added a batch metadata refresh flow with visible progress, per-item status badges, and the ability to stop an in-progress refresh.
  * Added support for refreshing metadata on selected items while keeping manual selections unchanged.
* **Bug Fixes**
  * Demo mode now blocks additional metadata write actions.
  * Improved matching for items with incomplete artist information by deriving details from the filename when needed.
  * Cancellation now stops refreshes cleanly and avoids starting with stale cancel state.

### DIFF
--- a/server.py
+++ b/server.py
@@ -240,6 +240,8 @@ _DEMO_BLOCKED: list[tuple[str, re.Pattern]] = [
     # anonymous demo visitors (they'd spend the shared rate limit).
     ("POST",   re.compile(r"^/api/enrichment/review/.+$")),
     ("POST",   re.compile(r"^/api/enrichment/kick$")),
+    ("POST",   re.compile(r"^/api/enrichment/cancel$")),
+    ("POST",   re.compile(r"^/api/enrichment/rematch$")),
     ("GET",    re.compile(r"^/api/enrichment/search$")),
     # Context menus (R2): the per-song re-match mutates the cache + spends
     # rate limit; Get-info exposes filesystem paths.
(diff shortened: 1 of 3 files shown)


byrongamatos:
Review pass (Claude Code). Solid. Backend: `rematch`/`states` cap the batch at 500 + string-filter inputs and look songs up by DB key (`enrichment_song_row`), so a crafted filename can't traverse the FS — an unknown name is simply skipped; `rematch` resets to `unscanned` with `allow_manual_overwrite=False` so user pins are preserved. The Stop button's cooperative cancel is a thread-safe `Event` checked between songs in both the match and art phases, and `_kick_enrich` clears it before launching so a stale cancel can't abort a fresh pass. The filename-artist seed (`_artist_title_from_filename`) is search-seed-only — displayed values still come from the confirmed match, and it returns None for non-CDLC names. Frontend badge poll loops are bounded (tick caps, stop on pass-end), matching the existing `_watchScan` pattern. 18 plumbing tests pass; node --check clean. CI red is the unrelated main-level starter-content bug. No change needed. (Codex second-opinion unavailable this session — rate limit — manual + test-backed.)

PR #766, opened 2026-07-04
fix(library): serve art/load for songs mounted through a library junction

ChrisBeWithYou:
## Symptom
Album art shows as **broken images** (and songs won't load) for any part of the library mounted through a directory **junction / symlink** — e.g. a library shared across app installs, or the desktop app's own mounts.

## Cause
The scanner (`dlc.rglob`) follows a junction and indexes the songs, so the grid shows them. But every path-validated handler goes through `_resolve_dlc_path` → `safe_join`, which does `(root / name).resolve()` and rejects anything whose **resolved** path leaves `DLC_DIR`. A junction resolves to its real target *outside* `DLC_DIR`, so every song reached through it is rejected → **403 `/art`**, **404 `/art/candidates`**, broken covers; `/load` breaks the same way.

## Fix
- **`_resolve_dlc_path`** now uses **lexical containment** (`os.path.normpath`, no symlink following), so an in-library junction is allowed while `..` traversal and absolute paths are still rejected. The traversal contract (`../` → 403, safe-but-missing → 404) is unchanged and pinned by the existing tests.
- **`safe_join` stays strict** (`.resolve()`-based) — it's the **zip-slip / plugin-asset / avatar** guard, where following a symlink *out* is the desired defense, so it is deliberately **not** loosened. It does gain an explicit **NUL guard**: on Python 3.13 (Windows) `resolve()` no longer raises `ValueError` on an embedded NUL, so the byte was leaking through containment; the guard is strictly-more-rejection with no effect on the zip-slip contract (fixes a pre-existing red test on 3.13/Windows).

## Security
Network-facing traversal (`..`, absolute, NUL, URL-encoded) stays fully blocked (unit + endpoint tests). The only new capability is *following a symlink that already lives inside the user's own library* — which requires local write access to plant (and all write routes are demo-blocked in hosted mode), while the loose-folder art handler keeps its own per-file `relative_to(song_path)` symlink re-check. Manual review done; the higher-risk callers retain strict `safe_join`.

## Tests
`tests/test_dlc_junction.py`: junction subfolder allowed (+ resolves through to the file); `..`/absolute/NUL rejected; and a contrast assertion that `safe_join` still rejects the junction (stays strict). Existing `test_loose_traversal` / `test_safepath` / `test_art_candidates` stay green.

Verified live: junctioned city-pop packs that returned 403/404 now serve their real embedded JPEG covers (`/art` → 200 `image/jpeg`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path handling to more reliably reject unsafe inputs, including embedded NUL bytes, empty paths, absolute paths, and directory-traversal attempts.
  * Strengthened DLC path containment checks to ensure requested content stays within the expected DLC root, using stricter normalization rules.
  * Preserved access to valid in-DLC content even when reached through trusted library junctions/symlinks.
* **Tests**
  * Added coverage for DLC junction and path-containment edge cases, including Windows path variations.

### DIFF
--- a/lib/safepath.py
+++ b/lib/safepath.py
@@ -26,6 +26,13 @@ def safe_join(root: Path, name: str) -> Path | None:
     """
     if not name:
         return None
+    # Reject embedded NULs explicitly. This used to ride on `.resolve()`
+    # raising ValueError, but on Python 3.13 (Windows) resolve() no longer
+    # raises for an embedded NUL, so the byte would otherwise leak through
+    # containment. An explicit guard is strictly-more-rejection (no effect on
+    # the zip-slip / traversal contract).
+    if "\x00" in name:
+        return None
     safe = name.replace("\\", "/")
     try:
         root_resolved = root.resolve()
--- a/server.py
+++ b/server.py
@@ -5253,6 +5253,16 @@ def _resolve_dlc_path(dlc: Path, filename: str) -> Path | None:
     safe = filename.replace("\\", "/")
     if "\x00" in safe:
         return None
+    # Reject drive-letter / absolute paths in BOTH conventions. A POSIX "/x" is
+    # caught by the containment check below (the `/` operator discards `root`),
+    # but a Windows drive-absolute "C:/x" is treated as a relative "C:" dir on
+    # POSIX and would otherwise slip in as `<root>/C:/x` — so the contract must
+    # hold cross-platform (a shared library is reached from either OS).
+    from pathlib import PurePosixPath, PureWindowsPath
+    if (PurePosixPath(safe).is_absolute()
+            or PureWindowsPath(safe).is_absolute()
+            or PureWindowsPath(safe).drive):
+        return None
     try:
         root = dlc.resolve()
         # normpath collapses `.`/`..`/duplicate separators purely lexically —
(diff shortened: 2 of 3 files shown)


byrongamatos:
Review + fix pass (Claude Code). The core security tradeoff is sound and well-documented: `_resolve_dlc_path` (the user's own trusted library) uses **lexical** containment so an intentional in-library junction works, while `safe_join` (zip-slip / plugin-asset / avatar — the untrusted surfaces, where following a symlink *out* is the defense) stays strict + gains the NUL guard. Consistent with the single-trusted-user model. Verified `..`/absolute/NUL are all still rejected.

**Pushed one fix (was red on the Linux CI):** your new `test_absolute_path_rejected` pins `C:/Windows/system32/x → None`, but on POSIX a drive-letter path isn't absolute, so `Path(dlc)/'C:/…'` became the *contained* dir `<dlc>/C:/…` and slipped through lexical containment (not an escape, but the test failed). Added an explicit reject for paths that are absolute or drive-qualified in **either** POSIX or Windows semantics (a shared library is reached from either OS) — relative/junction paths unaffected. `test_dlc_junction.py` now 5/5 green; the remaining CI red is the unrelated main-level starter-content bug. (Codex second-opinion unavailable this session — rate limit — manual + test-backed.)

PR #771, opened 2026-07-04
feat(enrichment): loose MusicBrainz search fallback (find aliased/romanized artists)

ChrisBeWithYou:
## Problem
The MB text search is a **strict field-phrase** query: `recording:"<title>" AND artist:"<artist>"`. A field phrase only matches MusicBrainz's **primary** artist/title — it **never searches aliases**. So a recording stored under a non-Latin primary name (大橋純子) whose romanized form ("Junko Ohashi") is only an **alias** returns **zero results**, even though MB has the record. Large parts of a community library (romanized J-pop / city-pop charts) were effectively unsearchable — searches just came up empty.

## Fix — strict-then-loose fallback
- `build_recording_query(..., loose=True)` drops the field scoping and phrases for plain AND-ed term groups (`(telephone number) AND (junko ohashi)`), which searches the whole document **including aliases**.
- `_mb_search_recordings` runs the **strict** query first (unchanged, high precision); only when it returns **nothing** does it retry once with the **loose** query. Mainstream matches are untouched, and the extra throttled request is spent only on a miss. Candidates are re-scored by `rank_candidates`, so recall rises without lowering match quality (auto-accept still requires the per-field floors).

Benefits both the background Refresh-Metadata pass and the manual Fix-match search box.

## Verified live
| search (all **0** under the strict query) | after |
|---|---|
| Junko Ohashi / Telephone Number | 3 — incl. **大橋純子 — Telephone Number** ✓ |
| Anri / Windy Summer | 4 — incl. Windy Summer (remaster) ✓ |
| AC/DC / Highway to Hell | 4 @ score 1.0 — strict still hits, **no loose retry** (no regression) |

## Tests
- `build_recording_query` loose form (drops field phrases; missing-artist case).
- `_mb_search_recordings` retries loose on an empty strict result, and does **not** retry when strict hits (no wasted request).

## Follow-up (separate)
Alias-aware **scoring** — compare the reference against a candidate's romanized aliases — so these can **auto-confirm**, not just appear as manual candidates (today the artist-script mismatch keeps the auto score low).

## Summary by CodeRabbit

* **New Features**
  * MusicBrainz recording searches now automatically retry with a broader “loose” match when the initial strict search finds no candidates.
  * Query matching supports a flexible mode for wider title/artist alias detection.

* **Bug Fixes**
  * Improves recording lookup accuracy, including scenarios where strict query formatting would otherwise prevent results.

* **Tests**
  * Added unit and server-level tests covering strict-to-loose retry behavior and verifying live-only exclusion rules.


byrongamatos:
Review + fix pass (Claude Code). The loose-fallback logic is correct: `denoise()` strips every Lucene special (`/ + - ! ( ) [ ] :`) before the term-group query, so the unescaped `(telephone number) AND (junko ohashi)` is injection-safe while CJK primaries survive; `_mb_search_recordings` only retries loose on an empty strict result and skips the retry when the loose query equals the strict one (no wasted request).

**Fixed:** this branch predated #758 (now merged to main), so it modified the *old* `build_recording_query`/`_mb_search_recordings`. Merged current main in and resolved the conflicts by **integrating** the loose fallback onto #758's versions — keeping its `-secondarytype:Live` filter (live-chart-gated) + `limit=12`, and updating the stale strict-query test to expect the live filter. 55 mb tests green; now conflict-free vs main. (Codex second-opinion still rate-limited — manual + test-backed.)

PR #772, opened 2026-07-04
feat(enrichment): alias-aware scoring — auto-confirm non-Latin-primary artists

ChrisBeWithYou:
**Stacks on #771** (base = `feat/mb-loose-search-fallback`; retarget to `main` once #771 merges).

## Problem
The loose-search fallback (#771) makes a recording stored under a **Japanese primary name** (大橋純子) *findable* via its romanized alias. But the **scorer** still compared the reference ("Junko Ohashi") against the candidate's **primary name only** → artist similarity `0` → below `AUTO_ARTIST_MIN`, so these could only ever appear as manual candidates (score ~0.5), never **auto-fill** on the Refresh pass.

## Fix
- **`mb_match.cand_artist_sim`** — best similarity over the candidate's primary name **and** its `artist_aliases` (romanized/alternate names). `score_candidate` and `classify` both use it, so an alias match clears the per-field artist floor.
- **`_mb_artist_aliases(id)`** — fetches an artist's aliases via `/ws/2/artist/{id}?inc=aliases`, **process-cached** (a one-artist discography costs a single request). Raises on transport failure so a pass pauses cleanly (nothing cached on failure).
- **`_alias_enrich`** — attaches aliases **only to promising near-misses** (title already agrees, primary artist doesn't), bounded by `_ALIAS_ENRICH_MAX=3`. A normal pass (artist already matches, or no ref artist) spends **zero** extra requests. Wired into both the auto-matcher (`_enrich_one`) and the manual search proxy (ranking).

## Verified live
| search | loose-only (#771) | + alias scoring |
|---|---|---|
| Junko Ohashi / Telephone Number → 大橋純子 | score **0.5** (manual only) | score **1.0**, ranked #1, **auto-confirmable** |
| AC/DC / Highway to Hell | 1.0 | 1.0, **no alias lookup** (unchanged) |

## Tests
`cand_artist_sim` uses aliases; an alias lifts a title-matching candidate from non-auto to `auto`; `_mb_artist_aliases` fetches + caches (+ rejects a non-UUID id without fetching); and an end-to-end `_background_enrich` auto-matches a Japanese-primary candidate through the alias.

## Note / possible follow-up
The stored canonical artist is MB's **primary** (大橋純子). If preferring the romanized display is wanted, a small display-precedence tweak (prefer a romanized alias / the pack's own value) can follow — the match + album/year/art/MBID all resolve correctly regardless.

## Summary by CodeRabbit

* **New Features**
  * MusicBrainz matching now recognizes alternate artist names, including romanized and other aliases, improving match quality for multilingual releases.
  * Manual search results are ranked using these alias-aware matches for better candidate suggestions.

* **Bug Fixes**
  * Improved auto-detection and review gating for releases where the track title matches but the artist appears under a different spelling or script.
  * Reduced missed matches for recordings with equivalent artist names across languages.


byrongamatos:
Review + fix pass (Claude Code) — stacked on #771. Logic is correct: `cand_artist_sim` maxes over primary + aliases (early-break at 1.0); `_mb_artist_aliases` is process-cached, UUID-guarded, and **raises** on a network failure so nothing is cached on failure (verified `_mb_http_get` raises for network-shaped errors, returns None only on 404); `_alias_enrich` only enriches title-matching / primary-artist-missing near-misses, bounded to `_ALIAS_ENRICH_MAX`; and the auto-matcher's propagated `EnrichTransportError` rides the existing `try/except → break` around `_enrich_one` (pauses the pass cleanly), while the manual search swallows it as a nicety.

**Fixed (integration):** synced the updated #771 base (which now carries merged #758 + #764) and resolved the server.py conflicts — routing `_enrich_one`'s search/enrich/rank through `ref` instead of `row` so alias scoring also uses **#764's filename-artist seed**, and keeping **#758's** duration corroboration in the manual search. 60 mb tests green. Still needs a retarget to `main` once #771 lands (as your PR body notes). (Codex second-opinion rate-limited — manual + test-backed.)

PR #777, opened 2026-07-05
feat(v3 library): "Fix metadata" popup — per-song override + lock, cover picker, MusicBrainz + AcoustID

ChrisBeWithYou:
A Plex-style **"Fix metadata"** editor for the v3 library, reached from a card's right-click / ⋮ → **Fix metadata…**. It's the manual curation tool for packs whose metadata is wrong or blank (a large slice of community/CDLC charts) — the ones automatic matching can't crack.

Everything here is a **reversible display layer — pack/`.sloppak` files are never modified.**

### Details tab — type + lock the shown fields
- Edit **title / artist / album / year**; each cell shows **Yours/Pack** provenance, a **lock** (pins the field so an auto-match can't re-canonicalize it), and **revert-to-pack**.
- Backed by a new per-field override + lock store (`song_field_override`): filename-keyed so it survives a rescan, dropped only with the song. `GET`/`PUT /api/song/{fn}/overrides` (GET also returns the `pack` baseline for provenance/revert).
- The lock is enforced server-side: the matcher composes a per-song lock filter (`_compose_lock_filter`) so an auto-match still fixes identity/art but won't overwrite a locked display field, and gap-fill skips locked fields. An explicit user pick always wins.

### Grid shows your overrides
- `query_page` overlays the override (over the pack value, and over the P4 artist-alias re-label) for the visible window — **matched MusicBrainz canon never silently re-titles a card** (canon stays in the drawer + art). Keyset-safe: paging still seeks on the raw column (the one overridable keyset column, title, stashes its raw value for the cursor; stripped from the payload).

### Cover art tab
- Hands off to the shared cover picker (Cover Art Archive / paste a link / upload); the pick refreshes the art everywhere.

### Match tab — MusicBrainz + Identify by audio
- The existing MB search/candidate/pick flow, refactored into shared body/footer helpers (the queue-review flow is untouched).
- **Identify by audio** (AcoustID, builds on #759): fingerprints the pack's own master audio when text search can't tell the studio take from live/comp versions. Outcomes are explicit — ✓ fingerprinted-with-matches / ✓ fingerprinted-but-no-match / no-audio / off — so an empty result reads as "ran, found nothing" (with a pointer to the manual tabs), never as a broken button.

### Notes
- Card menu item relabeled **"Fix metadata…"** (was "Fix match…"); replaces the thin single-song fix-match modal with the tabbed shell.
- `tailwind.min.css` regenerated for the popup's new utility classes.
[...]

PR #778, opened 2026-07-05
feat(v3 library): "Write to file" in the Fix-metadata popup

ChrisBeWithYou:
Completes the confirmed edit model for the Fix-metadata popup (stacked on #777).

Today the popup's **Save** keeps edits as a **reversible display overlay** — the pack/`.sloppak` files are never touched. This adds the deliberate opt-in counterpart: **Write to file** bakes the shown **title / artist / album / year** into the pack itself, so other tools/apps see the corrected tags and the fix survives a full rescan.

### How
- Reuses the **existing** `POST /api/song/{fn}/meta` (already on `main`: writes the manifest via `songmeta.write_song_metadata`, re-stats, coalesces a rescan). **No new backend** — and no dependency on the heavier #733 overwrite-with-backups feature, which is a separate thing.
- On a real file write (`persisted: true`): the now-redundant override **values** are cleared (any **locks** are kept), and the Details tab re-renders — so the fields read from the file as **Pack**, honestly reflecting that the fix now lives in the pack, not just the overlay.
- Loose-folder / unwritable packs fall back to the endpoint's DB-only update; the UI says so plainly ("may revert on a full rescan") and keeps the overlay.

### UX
- A secondary **Write to file** button beside the primary **Save** (Save stays the default, non-destructive path — matches the "files untouched unless you ask" ethos). Tooltip spells out that it's permanent and touches only the four fields, leaving the rest of the pack verbatim.

### Testing
- `node --check` clean. Verified on the testbed against the real library: the write path returns `persisted: true` (repacks the `.feedpak`); Save vs Write-to-file are distinct, and a persisted write flips the field provenance from **Yours** to **Pack**.

Stacked on **#777** — merge that first (or land them together).

## Summary by CodeRabbit

* **New Features**
  * Added a **“Write to file”** option in the Details tab for metadata edits.
  * Users can now permanently apply shown title, artist, album, and year values to the stored song file, in addition to the existing reversible save behavior.

* **Bug Fixes**
  * Improved status messaging to clearly distinguish between temporary overlay changes and changes written to the file.
  * After a successful write, the display now refreshes to reflect the updated stored metadata while preserving field locks.

### DIFF
--- a/static/v3/match-review.js
+++ b/static/v3/match-review.js
@@ -608,8 +608,16 @@
             if (status) { status.className = 'text-xs leading-relaxed text-fb-accent'; status.textContent = 'Could not write to the file — try again.'; }
             return;
         }
-        // Keep the in-memory song + grid in step with what was written.
-        for (const [f] of DETAIL_FIELDS) song[f] = fields[f];
+        // Keep the in-memory song + grid in step with what was *persisted*, not
+        // the raw input: the server coerces a non-numeric/empty year to "" (see
+        // update_song_meta), so mirror that here or the grid card flashes the
+        // typed text (e.g. "abcd") until the next natural refresh corrects it.
+        const applied = { ...fields };
+        if ('year' in applied) {
+            const yr = /^[+-]?\d+$/.test(applied.year) ? parseInt(applied.year, 10) : 0;
+            applied.year = yr ? String(yr) : '';
+        }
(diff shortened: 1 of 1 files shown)


ChrisBeWithYou:

1. #777 — Fix-metadata popup
2. #778 — Write to file
4. #780 — Genre in Details

PR #779, opened 2026-07-05
feat(v3 library): Match→Details "use these values" bridge

ChrisBeWithYou:
Connects the Fix-metadata popup's two tabs (stacked on #778 → #777).

**The gap:** the **Match** tab finds the correct recording (MusicBrainz search or Identify-by-audio), but by design a match only improves the *underlying* canon + art — it never silently re-titles the grid ("author-set shows by default"). So to actually *display* the match's values you had to retype them into **Details** by hand.

**The bridge:** each Match candidate now shows a **"Use these values →"** action. Clicking it:
- copies the candidate's **title / artist / album / year** into the Details tab as **pending, unsaved** inputs (shown pre-filled with a "review, then Save or Write to file" nudge),
- **pins the match** too (so the art/canon follow),
- and drops you on **Details** to review.

You stay in control — the values are populated for review, not force-saved, so you can tweak (e.g. prefer romaji over kanji) before **Save** (overlay) or **Write to file** (#778). The "never silently adopt a match" rule is intact; this is just the explicit one-click path to opt in.

Queue-review candidates are untouched — they still accept/pin on click.

### Testing
- `node --check` clean; live on the testbed. Verified the served bundle wires `useTheseValues` at both candidate sites (search + Identify).

Stacks on **#778** (Write to file) → **#777** (popup). Merge bottom-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Single-song match selection now fills the Details tab instead of immediately finalizing the choice, giving you a chance to review and edit values first.
  * The interface now switches to the Details tab after a candidate is selected and shows a status message confirming the fields were populated.

* **User Experience**
  * Candidate rows now include a clearer hint indicating the selected values will be used in the Details form.

PR #780, opened 2026-07-05
feat(v3 library): genre field in the Fix-metadata popup

ChrisBeWithYou:
Adds **Genre** as a fifth field in the popup's Details tab (stacked on #779 → #778 → #777).

### What
- Genre gets the same treatment as the other fields — edit, **lock** (pin against auto-match), **revert**, and Yours/Pack provenance — backed by the existing override store (which already allowed `genre`).
- **Made it actually useful:** the genre **filter** and **facet** (`/api/library/genres` + the `genre=` filter) now resolve the per-song override — *effective genre = your override else the scanned pack genre* — so a corrected or added genre is immediately **browsable**. Guarded on `_has_genre_overrides()` so the common no-override case stays on the plain indexed column (zero cost).
- **Genre stays a library-only overlay** — it is deliberately **not** a write-to-file field. `WRITE_FIELDS` (the four file-safe fields) is split from the five `DETAIL_FIELDS`, so **Write to file** writes title/artist/album/year and leaves the genre override in place; the copy says "genre stays a library-only tag."
- The Match→Details bridge (#779) also carries a candidate's first genre into the field.

### Why genre isn't written to the pack
`POST /api/song/{fn}/meta` (the write path) handles the four catalog fields; the feedpak `genres` list is a separate shape and genre works best as a fast, reversible library-side facet (fix it, browse by it, never touch the file). Locking still protects it from auto-match.

### Testing
- New tests: the effective-genre facet + filter surface/return an overridden song; a value-less genre **lock** doesn't invent an effective genre. Full `test_field_overrides` + `test_library_filters` green (57).
- Verified live on the testbed: overriding a city-pop pack's genre to "City Pop" makes it appear in the facet and filterable (1 song matched), cleared cleanly.

Stacks on **#779 → #778 → #777**. Merge bottom-up.

## Summary by CodeRabbit

* **New Features**
  * Genre overrides now count in library browsing and filtering, so the app shows the effective genre instead of only the stored file value.
  * The metadata review screen now displays Genre in the pending details view.
* **Bug Fixes**
  * Writing metadata now preserves genre overrides instead of clearing them.
  * Genre lookups and filters now stay consistent when a song has an overridden genre value.

### DIFF
--- a/server.py
+++ b/server.py
@@ -1277,6 +1277,27 @@ class MetadataDB:
             (filename,)).fetchone()
         return {k: ((row[i] or "") if row else "") for i, k in enumerate(keys)}
 
+    # Effective genre = a per-song genre OVERRIDE (Fix-metadata popup) else the
+    # scanned pack genre. Applied at FILTER/FACET time (like the P4 artist alias)
+    # so a corrected genre is browsable — the correlated subquery is used ONLY
+    # when genre overrides actually exist; the common case stays on the plain
+    # indexed `genre` column. Genre stays a library-only overlay (it isn't a
+    # write-to-file field), so it never touches the pack.
+    _EFFECTIVE_GENRE_SQL = (
+        "COALESCE((SELECT o.value FROM song_field_override o "
+        "WHERE o.filename = songs.filename AND o.field = 'genre' "
+        "AND o.value IS NOT NULL AND o.value != ''), genre)"
+    )
+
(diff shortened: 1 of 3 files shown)


ChrisBeWithYou:

1. #777 — Fix-metadata popup
2. #778 — Write to file
4. #780 — Genre in Details

PR #781, opened 2026-07-05
feat(v3 library): persistent "no match" badge + Unmatched quick filter

ChrisBeWithYou:
Makes the metadata **misses visible and reachable** — the two follow-ups to the Refresh-Metadata batch (#764). Independent of the Fix-metadata popup stack; branches off `main`.

**The gap:** the batch paints a per-tile "no match" only *while a pass runs*, then clears it — so at rest you can't see which songs (the ~640 city-pop imports) never matched.

### Persistent per-card "No match" badge
- `query_page` now marks each row `unmatched` — a cheap `match_state='failed'` set-membership over the page (like `favs`/`estd`), so no N+1.
- `enrichBadge` paints a **subtle resting marker** on those cards (right-click → fix by hand). It's tracked in a `_unmatched` set so a batch tile clearing falls back to the resting badge instead of wiping it; a **live batch tile still wins** while a pass runs (they never coexist).

### "Unmatched" quick filter
- A local-only **Unmatched** toolbar toggle: one click applies the same filter as the drawer's *Match → Unmatched* (`match_state='failed'`), re-queries, and reflects active state — so the no-match pile is one click from the batch, without digging into the filter drawer.

### Testing
- `query_page` flags a failed row (`unmatched: true`) and not a matched one; the `match=unmatched` filter returns only the failed song. `test_library_filters` + keyset + enrichment suites green (67). `node --check` clean; reuses existing utility classes (no CSS regen).

## Summary by CodeRabbit

* **New Features**
  * Added persistent **No match** (unmatched) badges on song cards for items that failed metadata matching.
  * Added an **Unmatched** quick filter to show only songs with no metadata match.

* **Bug Fixes**
  * Song cards retain the no-match indicator across metadata refreshes, instead of clearing when temporary batch tile state is gone.

* **Tests**
  * Added coverage for the unmatched flag behavior and the **match=unmatched** quick filter results.

### DIFF
--- a/server.py
+++ b/server.py
@@ -3038,6 +3038,22 @@ class MetadataDB:
                     out[fn] = st
         return out
 
+    def _unmatched_set(self, filenames) -> set:
+        """The subset of `filenames` whose enrichment landed in the 'failed'
+        (no-match) state — feeds the grid's persistent per-card "no match" badge,
+        so the misses stay visible at rest (the batch tile only shows while a
+        refresh runs). Chunked set membership, like favorite_set."""
+        fns = list(filenames)
+        out: set = set()
+        for i in range(0, len(fns), 400):
+            chunk = fns[i:i + 400]
+            if not chunk:
+                break
+            q = ("SELECT filename FROM song_enrichment WHERE match_state = 'failed' "
+                 "AND filename IN (%s)" % ",".join("?" * len(chunk)))
+            out.update(r[0] for r in self.conn.execute(q, chunk).fetchall())
+        return out
+
     def enrichment_song_row(self, filename: str) -> dict | None:
         """The identity fields the matcher/scorer keys on, for one song."""
         row = self.conn.execute(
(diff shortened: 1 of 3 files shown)


ChrisBeWithYou:

1. #777 — Fix-metadata popup
2. #778 — Write to file
4. #780 — Genre in Details

ChrisBeWithYou:
**Integration note** — from a full-queue dry-run that merged every open org PR into a next-version test build.

Your persistent "No match" badge now overlaps **#783** (the Fix-metadata capstone), which turns the same badge into a **clickable** hand-off into the Fix-metadata popup (`data-meta-fix`/`fixable` in `songs.js`). When both are merged, #783's clickable badge is the superset and #781's non-interactive badge is redundant.

**Recommend: rebase on #783** and trim this PR down to just its unique surface — the **Unmatched quick-filter** button + `toggleUnmatchedFilter` + the Match filter section. Those merge cleanly and are the real contribution; the badge itself is already covered by #783. (In the dry-run I kept #783's clickable badge, and #781's filter + toggle slotted in intact.)

Issue #111, opened 2026-06-20
Adding a vst plugins crashes [Gojira X]


byrongamatos:
> _Restored from `slopsmith/slopsmith#280` — original issue, opened by @danom21 on 2026-05-15._
> _[restored-from: slopsmith/slopsmith#280]_

Every time I try to add a vst plugin to my signal chain my game closes

byrongamatos:
_@Durrgon commented on 2026-05-15:_

Also having the same issue, but for some reason if I set my saved signal chain as default using the VST it works fine.

Reproduction steps:
Plugin -> Audio -> Signal chain section -> Purple VST button -> Add VST -> Try to play any song and it will crash

However, if you do the above, and then go to the presets section, and press Save Current Chain, then go to Settings -> Audio Engine -> Saved presets section, and set it as default it seems to work correctly.

Logs I found didn't have anything useful in them

byrongamatos:
_@byrongamatos commented on 2026-05-16:_

Do you have trial plugins that have expired? Please redownload the 0.2.8 prerelease it was updated this morning and try again

byrongamatos:
_@danom21 commented on 2026-05-17:_

> Do you have trial plugins that have expired? Please redownload the 0.2.8 prerelease it was updated this morning and try again

No I have the full versions of neuraldsp gojira x and thallamp, I was having the scan plugins crashing issue others were. I updated and its not crashing after searching anymore, but now I'm having an issue where when I try to open them the game freezes, says "not responding" for a few seconds, then comes back with the vst not open. I got thallamp to open once but the game crashed after trying to save the tone.

byrongamatos:
_@Hexocutioner commented on 2026-05-18:_

i have found success as long as i close the vst window before saving the chain 
 https://youtu.be/-vNE2k73Eac

byrongamatos:
_@Durrgon commented on 2026-05-18:_

> i have found success as long as i close the vst window before saving the chain https://youtu.be/-vNE2k73Eac

This seems to be what's happening with me as well

byrongamatos:
_@Hexocutioner commented on 2026-05-18:_

I have also found that if I separate the stems to turn down the guitar
that with a vest running I cannot turn down the stems also I cannot turn
down the mix between my guitar and the song either

byrongamatos:
_@mogul commented on 2026-06-08:_

Please check again with -beta.2. We've made lots of improvements and observe no issues except with Gojira X at this point.

PR #808, opened 2026-07-07
feat(drums): capture velocities alongside times in unmapped-percussion reporting

ChrisBeWithYou:
Additive companion to **feedBack-plugin-editor#111** (drum velocity authoring): the editor''s unmapped-notes dialog can now carry real dynamics through hand-mapping instead of flattening every note to `v:100` — but only if the converters capture them. This PR adds that capture.

## What
- Both drum converters'' opt-in `out_unmapped` reporting — `convert_drum_track_from_midi` (`lib/midi_import.py`) and `convert_drum_track_to_drumtab` (`lib/gp2rs.py`) — gain an **index-aligned `velocities` list** next to `times`:
  - MIDI path: `msg.velocity` verbatim (already in hand at the capture site — zero extra work on the default no-capture path).
  - GP path: the note''s velocity with the **same 1–127 gate as mapped hits**, falling back to the 100 import default for corrupt/zero values.
- The GP path''s end-of-conversion chronological sort now reorders `times` and `velocities` **in lockstep** (pair sort) — multi-voice measures capture out of order, and sorting times alone would silently reassign dynamics to the wrong notes.
- Fully additive: the dict shape gains one key; callers that ignore it (older editor, any other consumer) are byte-for-byte unaffected. The editor''s response shaping (#111) passes it through when present and omits it cleanly when absent.

## Tests
Extended `tests/test_midi_import_drums.py` + `tests/test_gp2rs_drums.py`: alignment with times, the lockstep sort under reversed chronology, and the out-of-range fallback — 26 passing. (9 pre-existing local-env failures — tailwind rebuild + plugin-loader e2e — reproduce identically on pristine main and are untouched by this diff.)

https://claude.ai/code/session_01JEoFeTPSnz4NpwwCG52hnu

## Summary by CodeRabbit

* **New Features**
  * Opt-in unmapped percussion import/export now records per-hit MIDI `velocities` alongside `times` for both MIDI and Guitar Pro conversions; existing consumers ignoring the field are unaffected.
* **Bug Fixes**
  * Preserves time/velocity correspondence when hits are re-sorted chronologically; when velocity is missing/corrupt (e.g., 0), it defaults to `100`.
* **Tests**
  * Expanded coverage to verify velocity alignment, correct reordering of `times`+`velocities` together, and defaulting behavior.


byrongamatos:
@coderabbitai review

PR #825, opened 2026-07-09
fix(keys_highway_3d): stop auto-connect clobbering the global MIDI device

gionnibgud:
## Problem

Opening the **Keys Highway 3D** plugin could silently switch the user's configured MIDI device. Reported symptom: the highway attaches to the wrong keyboard; disconnecting every other MIDI device made it finally pick the intended one.

Root cause was two coupled defects in the plugin's MIDI selection:

1. **`_midiAutoConnect` ignored the global selection.** It only consulted the plugin's own `localStorage` pick (`keys3d_midi_pick`); with none saved it fell straight through to *"first non-loopback device"*, never reading the core `midi-input` domain's global selection (Settings → Input Setup, `window.slopsmith.midiInput.getSelected()`).
2. **`_midiConnect` persisted every connect unconditionally** — to both the local pick *and* the shared domain selection (`mi.select`). So the first-device guess got frozen locally **and overwrote the global default** that other consumers (drums, Input Setup) rely on.

## Fix

- **Domain-wide selection is now the source of truth.** `_pickMidiTarget` resolves `global → legacy local pick (fallback + name-recovery for stale ids) → first device`. A stale local pick can no longer override the globally-configured device. The hotplug-recovery guard now preserves *any* configured preference (global or local) rather than grabbing a random present device on a transient unplug.
- **Persistence is gated behind an explicit `persist` flag.** Only a deliberate device selection (`keysH3dSetMidiInput`) writes the local pick and the shared global. Auto-connect and programmatic audio-input opens (`_aiOpen`) open the resolved device for the session **without touching either store**, so opening this highway can't clobber the user's global default. `mi.select()` isn't needed to open (open takes the `logicalSourceKey` directly), so dropping it from the auto path costs nothing.

## Scope / follow-up

This makes keys honor the single global default without clobbering it. It intentionally does **not** add per-instrument routing — keys and drums still both follow the one global. The deeper fix (make the `midi-input` domain's selection instrument-scoped, so the already-per-instrument `input_setup` wizard actually persists per instrument) is core work tracked separately: #826.

## Tests

Pure decision logic extracted to `_pickMidiTarget` and covered by unit tests in `data_layer.test.js` (global-wins-over-stale-local, local fallback, id-regeneration name-recovery, blocklist handling, None opt-out, recovery cases). `node --test tests/*.test.js` → 39 pass.

## Summary by CodeRabbit

* **New Features**
  * Improved MIDI device auto-selection with a stricter priority order (global setting first, then local preference).
  * Enhanced hotplug recovery to avoid switching to an unintended device during transient disconnects.
* **Bug Fixes**
  * Loopback/blocklisted ports are now properly rejected; “None” reliably returns no selection.
  * User-initiated MIDI selections persist, while automatic/programmatic connections remain session-only.
* **Tests**
  * Expanded unit tests covering precedence, stale preference recovery, loopback handling, and recovery-mode behavior.
* **Chores**
  * Updated plugin version to **0.2.1**.

### DIFF
--- a/plugins/keys_highway_3d/plugin.json
+++ b/plugins/keys_highway_3d/plugin.json
@@ -1,7 +1,7 @@
 {
   "id": "keys_highway_3d",
   "name": "Keys Highway 3D",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "RS+-style 3D falling-note piano highway fed by the Sloppak Notation Format, with Web MIDI input scoring.",
   "type": "visualization",
   "bundled": true,
--- a/plugins/keys_highway_3d/screen.js
+++ b/plugins/keys_highway_3d/screen.js
@@ -813,20 +813,36 @@
         _writeStore(STORE_KEYS.midiPick, JSON.stringify({ id: id || '', name: name || '', key: key || '' }));
     }
 
-    function _midiAutoConnect(allowFallback) {
-        // Recovery (sources-changed after unplug) passes false: never switch to a
-        // fallback input, because _midiConnect persists the pick and that would
-        // overwrite the user's saved device on a transient multi-device unplug
-        // (the original returns on replug and reconnects then).
-        if (allowFallback === undefined) allowFallback = true;
-        const inputs = _midiSources();
-        if (!inputs.length) return;
-        const saved = _readSavedPick();
-        // Explicit "None" opt-out.
-        if (saved && saved.id === '' && saved.name === '') return;
-        // Prefer the globally-unique logicalSourceKey, then the legacy bare
-        // sourceId, then case-insensitive name (Chrome on Linux regenerates ids
-        // per page load), then first non-loopback.
+    // Pure decision logic (exported via __test): pick which device to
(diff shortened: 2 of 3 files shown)


gionnibgud:
Addressed both CodeRabbit findings in the amended commit:

1. **Loopback fallback** (`_pickMidiTarget`): the first-device fallback was `inputs.find(notBlocked) || inputs[0]`, which could still return a blocklisted "Midi Through"/IAC port when *every* candidate is a loopback. Changed to `|| null` — connect to nothing rather than a dead port that silently eats every note. Added a unit test for the all-loopback case.

2. **Global not persisted on the no-instance path** (`_midiConnect`): the `_instances.size === 0` early return fired *before* `mi.select`, so an explicit selection with no live renderer wrote the local pick but not the shared global — breaking the "explicit selection writes both" contract. Hoisted the `if (persist) await mi.select(lkey)` above the early return (best-effort, so a select hiccup can't abort the connect).

Tests: `node --test tests/*.test.js` → 60 pass.

PR #828, opened 2026-07-09
feat(audio): renderer-bus feeder — song audio into engine output under exclusive mode (Phase 2)

OmikronApex:
Stacked on #824 (Phase 1) — merge that first; this PR then reduces to one commit.

Under exclusive-style output the Phase 1 native transport carries loose `/audio/` songs and feedpak full-mixes, but not the stems plugin's multi-stem WebAudio graph or tracks JUCE hard-rejected. This adds `_installRendererBusFeeder` to `static/app.js`:

- Captures the renderer-side master (stems plugin's published `audioGraph`, or the core `<audio>` element as fallback) with an AudioWorklet tap.
- Re-points the owning AudioContext at a null sink (`setSinkId({type:'none'})`) so the graph keeps rendering while the device is exclusive-locked — analysers/visualizations stay live.
- Pushes ~10 ms chunks over IPC into the desktop engine's renderer bus (feedBack-desktop `feat/renderer-bus-phase2`), mixed like a backing track (~10–20 ms added latency on song audio only; guitar path untouched).
- Reuses #824's `window._juceOutputIsExclusive` predicate; inert in the Docker sphere and in shared mode (no double audio — covered by tests).

Validated by the fix12 tester spike (null-sink rendering works, clocks hold, drift → 0, zero overflow over 8 min) and shipped end-to-end in the fix13 tester build.

Companion PRs: feedBack-desktop `feat/renderer-bus-phase2` (engine-side bus), feedBack-plugin-stems `feat/publish-audio-graph` (graph publication).

Tests: 7 new sandbox tests (`tests/js/renderer_bus_feeder.test.js`); full JS suite 955/955; #824's python suite 19/19 on the merged result.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed silent audio in desktop exclusive-output modes by keeping playback active and properly routing audio to the system.
  * Improved handling when switching between exclusive and shared output, so audio now resumes and stops more reliably.
  * Added support for both live stem playback and song playback when the renderer needs to supply audio to the desktop engine.

### DIFF
--- a/server.py
+++ b/server.py
@@ -12941,53 +12941,57 @@ _extract_cache = {}  # filename -> (tmp_dir, song, timestamp)
 _extract_cache_lock = threading.Lock()
 
 
-@app.get("/api/sloppak/{filename:path}/file/{rel_path:path}")
-def serve_sloppak_file(filename: str, rel_path: str):
-    """Serve a file from inside a sloppak (stems, cover, etc.)."""
+def _resolve_sloppak_local_file(filename: str, rel_path: str):
+    """Resolve a file inside a sloppak to its on-disk path.
+
+    Applies the same containment guards as ``serve_sloppak_file``. Returns the
+    resolved ``Path`` on success, or an ``(error, status)`` tuple on failure so
+    callers can produce their endpoint-appropriate response.
+    """
     dlc = _get_dlc_dir()
     if not dlc:
-        return JSONResponse({"error": "not configured"}, 404)
-    # `filename` is an attacker-controlled `:path` param. Contain it under
-    # DLC_DIR before it reaches the resolver, which does a bare
-    # `dlc_root / filename`. Without this, `../../../etc` escapes the root
-    # and the rel_path guard below validates `target` against the already-
-    # escaped `src`, which trivially passes — yielding arbitrary file reads
-    # (e.g. /api/sloppak/../../../../etc/file/passwd). Mirrors the guard
-    # `get_song_art` applies to the same filename param.
+        return ("not configured", 404)
+    # `filename` is caller-controlled. Contain it under DLC_DIR before it
+    # reaches the resolver (see serve_sloppak_file for the traversal rationale).
     resolved = _resolve_dlc_path(dlc, filename)
     if resolved is None:
-        return JSONResponse({"error": "forbidden"}, 403)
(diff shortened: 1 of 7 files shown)


byrongamatos:
Reviewed — ship-ready, one retarget.

The `off/element/stems` state machine is sound: `_busy` reentrancy guard, `stemsGraphChanged` re-engage, and disengage-restores-sink are all correct and the 7 sandbox tests cover exactly the tricky transitions (exclusive engage, return-to-shared disengage, native-transport/`_juceMode` inertness, element fallback, graph-swap, engine-stop). Namespace fallback `(window.feedBack || window.slopsmith)?.stems?.audioGraph` matches the stems plugin (which publishes on `slopsmith`) — no skew. The permanent `createMediaElementSource` reroute is a real one-way side effect but acknowledged, and sink-toggling keeps it transparent in shared mode.

**Before merge: retarget to `main`.** Base is `feat/feedpak-exclusive-native-routing` (#824's branch), which is already merged — this should collapse to one commit against main. CodeRabbit skipped this PR for the same feature-branch reason, so it's had no bot review; the tests are the safety net and CI needs to actually run once it's on main.

Minor (optional): the tap worklet `postMessage`s every 128-frame quantum unconditionally — the `active` gate is on the main-thread receiver, so an initialized-but-inactive element tap keeps allocating + posting ~375×/s whenever `_elCtx` runs. Harmless, just avoidable; push the active flag into the worklet if it ever shows up in a profile.

OmikronApex:
Retargeted to `main` and synced — the diff is now exactly the feeder + its tests (`static/app.js` +203, `tests/js/renderer_bus_feeder.test.js` +195; the `main` merge picked up one post-#824 touch in `static/capabilities/audio-effects.js` as ancestry, not diff). Full JS suite 955/955 locally; CI should now run for real, and CodeRabbit can review against the default branch.

On the worklet nit (inactive tap still posting ~375×/s while `_elCtx` runs): agreed it's avoidable — pushing the `active` flag into the worklet via a port message is the fix. Deferring per your "if it ever shows up in a profile"; noted as a follow-up so it doesn't get lost. Worth remembering the exposure is bounded anyway: the element tap only exists after the first exclusive-mode engagement, and the stems tap's context is suspended whenever no song is loaded.

PR #833, opened 2026-07-10
feat(server): add appstate.py, the router seam (R3)

byrongamatos:
Third step of **R3**, after #830 (`MetadataDB`) and #831 (`AudioEffectsMappingDB`). This is the seam the router train needs — landed on its own, per the master plan's "`appstate.py` gets its own risk-managed PR, never folded into a neutral router split."

## The problem

Routes moving out of `server.py` need `meta_db` and friends. They must not `import server` — the graph goes circular the moment `server` imports them back.

## The shape

`server.py` keeps **constructing** its singletons and now **injects** them once:

```python
meta_db = MetadataDB(CONFIG_DIR)
appstate.configure(meta_db=meta_db, audio_effect_mappings=audio_effect_mappings)
```

and a router reads them back as module attributes, at call time:

```python
import appstate

@router.get("/api/artist/{name}/page")
def artist_page(name):
    return appstate.meta_db.artist_page(name)
```

Dependencies flow one way: `server → routers → appstate`. This is the Python analogue of the frontend refactor's injected `configureX({…})` seams (stems' `configureStreaming`, studio's `configureAudioGraph`, the editor's `src/host.js`) and of the plugin `setup(app, context)` contract in Principle III.

## Two properties that are load-bearing

**1. `import appstate` constructs nothing and touches no disk.** This is why `appstate` does *not* own the singletons, as originally sketched. **49 test files** do `sys.modules.pop("server")` + re-import to rebuild `meta_db` under a patched `CONFIG_DIR`. A singleton owned by `appstate` survives that pop and goes stale — reproduced directly:

```
first import        : DB@/tmp/A
pop server only     : DB@/tmp/A   <-- stale
pop server+appstate : DB@/tmp/C
```

With the injection seam, `server` remains the owner, so all 49 fixtures keep working **untouched** and the router still sees the fresh DB.

**2. Reads must be late-bound.** `from appstate import meta_db` freezes the binding, defeating both a later `configure()` and `monkeypatch.setattr(appstate, "meta_db", fake)`. Same read-only-binding trap as ES `import`. Routers must use `appstate.meta_db`. Pinned by a test.

## Guards

- `configure()` **raises on an unknown slot** rather than silently creating a global nothing reads. A seam whose wiring can no-op undetected is worse than no seam — the frontend refactor lost a regression to exactly that twice, when a scripted `setHostHooks` edit stopped matching its anchor.
- Defaults are `None`: inert but *type-honest*, so a router running before `configure()` fails loudly on `NoneType` instead of quietly operating on a stand-in.
- `tests/test_appstate.py` asserts `server` **actually calls** `configure`. **Negative-checked**: dropping the call fails exactly the two wiring tests while the other five stay green — those five are precisely the false-green a seam test must not be.

## Packaging (verified against the real Docker daemon)

`.dockerignore` opens with a blanket `*` exclusion and re-allows files one by one, so `COPY appstate.py /app/` **fails the image build** without an allowlist entry. Confirmed:

```
ERROR: failed to compute cache key: "/appstate.py": not found
```

Fixed with `!appstate.py`, then re-verified that the build context reaches `/app/appstate.py`. `docker-compose.yml` gains the dev bind-mount; `docker-compose.nas.yml` runs the baked image so the `COPY` covers it. **`routers/` will need both entries when it lands.**

## Verification

- `pytest` — **2348 passed** (2341 + 7 new), 4 skipped.
- `pyflakes` clean; `npm run lint` 0 errors.
- **Boot smoke**: `/api/version`, `/api/library`, `/api/audio-effects/mappings` all answer; all three migrated plugins still serve their `src/` graphs; `appstate.meta_db is server.meta_db` asserted against the live import.
- **Codex preflight**: found a real defect — the new suite imported `server` without patching `CONFIG_DIR`, so running the file alone created `web_library.db` + `audio_effects.db` in the developer's real `~/.local/share/feedback`. Reproduced, fixed with an `isolated_server` fixture (tmp_path + both connections closed on teardown), and the *full* suite now leaves the real config dir untouched. Re-reviewed: 0 findings.

## Summary by CodeRabbit

* **New Features**
  * Added shared application state wiring so server components can access configured database and audio mappings reliably.
  * Ensured fresh application state is published when the server is reloaded or reconfigured.

* **Bug Fixes**
  * Included the new application state module in Docker images and development containers.

* **Documentation**
  * Documented the application state integration and updated size-exemption guidance.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,10 @@
 !dockerfile
 !requirements.txt
 !server.py
+# The router seam server.py injects its singletons into (R3). Root-level Python
+# that ships in the image must be re-allowed explicitly — this file starts with
+# a blanket `*` exclusion. `routers/` needs the same treatment when it lands.
+!appstate.py
 !main.py
 !VERSION
 !tailwind.config.js
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`appstate.py` — the router seam (R3).** Route modules moving out of `server.py`
+  need `meta_db` and friends but must not `import server`, or the import graph goes
+  circular the moment `server` imports them back. So `server.py` keeps *constructing*
+  its singletons and now **injects** them once — `appstate.configure(meta_db=…,
+  audio_effect_mappings=…)` — and a router reads them back as module attributes at call
+  time (`import appstate; appstate.meta_db.…`). This is the Python analogue of the
+  frontend refactor's injected `configureX({…})` seams and of the plugin
+  `setup(app, context)` contract: dependencies flow one way, `server → routers →
+  appstate`. Two properties are load-bearing and pinned by `tests/test_appstate.py`:
+  (1) `import appstate` constructs nothing and touches no disk, so the ~49 test fixtures
+  that `sys.modules.pop("server")` + re-import (to rebuild `meta_db` under a patched
+  `CONFIG_DIR`) keep working untouched — a singleton *owned* by `appstate` would survive
+  that pop and go stale; (2) reads must be late-bound (`appstate.meta_db`, never
+  `from appstate import meta_db`), since a `from` import freezes the binding and defeats
+  both a later `configure()` and `monkeypatch.setattr` — the same read-only-binding trap
+  as ES `import`. `configure()` rejects an unknown slot rather than silently creating a
+  global nothing reads, and the suite asserts `server` actually calls it (a seam whose
+  wiring can no-op undetected is worse than no seam). Ships in the image via
+  `Dockerfile` `COPY appstate.py /app/` plus a `.dockerignore` allowlist entry — that
+  file opens with a blanket `*` exclusion, so root-level Python must be re-allowed
+  explicitly or the build fails.
+
 ### Changed
 - **`AudioEffectsMappingDB` moved out of `server.py` into `lib/audio_effects_db.py`
   (R3, move-only).** The core-owned song/tone → provider routing index follows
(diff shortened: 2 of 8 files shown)


byrongamatos:
Both CodeRabbit findings were real and are fixed in `7ba9d4a`.

**1. Stale slots on teardown** — accepted. The fixture closed `server`'s DB connections but left `appstate.meta_db` / `appstate.audio_effect_mappings` published and pointing at closed sqlite handles, i.e. a live-looking dead singleton for any later test. Teardown now snapshots both slots before the import and restores them after.

**2. The re-import test didn't re-import** — accepted, and the sharper of the two. I introduced that regression myself while fixing Codex's `CONFIG_DIR`-isolation finding: the test collapsed into re-asserting what `test_server_wires_the_seam` already covers, so it could not detect the staleness it names. It now pops `server`, re-imports under a **second** `CONFIG_DIR`, and asserts the seam republishes:

```python
assert second_server.meta_db is not first_db            # genuinely rebuilt
assert str(second_config) in second_server.meta_db.db_path
assert appstate.meta_db is second_server.meta_db        # ...and re-published
```

Both fixes are **negative-checked**, since a guard that passes either way guards nothing:

- Simulating an appstate-*owned* singleton (`configure()` made only-first-wins, which is what an owned module-level `meta_db` behaves like across a `sys.modules.pop("server")`) now **fails** `test_reimporting_server_republishes_the_fresh_singletons`. It did **not** fail against the previous version — exactly your point.
- Dropping `server`'s `appstate.configure(...)` call still fails precisely the two wiring tests, and no others.

One note on the committable suggestion: it placed the snapshot assignments *above* the fixture docstring, which would have demoted the docstring to a bare expression statement and dropped `__doc__`. Written by hand instead, same semantics.

`pytest` 2348 passed; the full suite leaves the real `~/.local/share/feedback` untouched. Codex re-review: 0 findings.

PR #837, opened 2026-07-10
Update proxmox build to match routers/appstate.py changes in Dockerfile

mbanks850:
## Summary by CodeRabbit

* **Bug Fixes**
  * Proxmox container builds now include router components and application state files required at runtime.
  * Builds fail clearly when essential application files are missing, preventing incomplete deployments.


mbanks850:
Cancelling, with the Routers and appstate.py move to /lib.   this pull request is not needed

PR #839, opened 2026-07-10
refactor(server): extract the loops routes into routers/loops.py (R3)

byrongamatos:
Third router in the R3 train (after audio-effects #834, artist-aliases #838). Move-only.

Practice loops — saved A/B regions per song. `router_scan.py` ranks it at **0 `setattr` targets, 0 helpers to relocate**; `meta_db`-only, own banner, contiguous.

- `GET /api/loops` · `POST /api/loops` · `DELETE /api/loops/{loop_id}`

Bodies verbatim; `@app` → `@router`, `meta_db` → `appstate.meta_db`. `include_router` at the original site. **`server.py`: 9,337 → 9,302.**

## Verification
- **Route table identical** to `origin/main` (143, paths/methods/order).
- No test retargeting (`test_demo_mode` only names the paths in middleware regexes).
- **Boot smoke**: `POST` auto-names `Loop N`, `GET` reads back, `DELETE` removes, missing fields → error; **demo mode `403`s both writes, allows the read**.
- `pytest` 2398 passed; packaging guard green; `pyflakes` clean; `npm run lint` 0 errors; **Codex 0 findings**.

Next candidates (`router_scan.py`): `wanted`/`saved` need `_clean_str` (shared by 10 callers) relocated to a `lib/` home first — a small shared-helpers decision before those move.

## Summary by CodeRabbit

- **New Features**
  - Added API support for listing, saving, and deleting audio loops.
  - Automatically assigns a default name when saving an unnamed loop.
  - Loop results are ordered by their start time.

- **Documentation**
  - Updated release notes and project size documentation to reflect router extraction progress.


byrongamatos:
Both findings are real, and both are **pre-existing** — the `COUNT(*)` sat outside `_lock` and `list_loops`' SELECT was unlocked on `origin/main` too (verified byte-identical to `6c98aba:server.py`). This PR moved the code verbatim, so they correctly weren't part of a move-only change. Fixing them here would have made the extraction non-verbatim.

Landing both in a small follow-up: move the `COUNT(*)` inside the `with _lock` block (atomic name generation) and wrap the list read in the lock. Severity is low in context — FeedBack is single-user (Principle I), so two concurrent unnamed-loop POSTs essentially can't happen — but the fix is one line each, so it's worth doing. Follow-up PR incoming; will link it here.

byrongamatos:
Follow-up fix opened: https://github.com/got-feedBack/feedBack/pull/840 — locks the count+insert and the list read, with a barrier-based concurrency test (negative-checked 5/5).

PR #841, opened 2026-07-10
refactor(server): extract the playlists routes into routers/playlists.py (R3)

byrongamatos:
The biggest router yet — 12 playlist routes + custom covers — and the first that needs the **config-path seam**. `server.py`: **9,302 → 9,085** (−217).

## Three causally-linked pieces (all required for playlists)

1. **`config_dir` joins the appstate seam.** The plan always put the path constants in appstate; I deferred them in S3 to dodge the `STATIC_DIR` retargets, and now need `config_dir` for playlist covers. It's **env-derived**, so the ~49 pop-and-reimport fixtures reconfigure it for free — **zero `setattr` retargeting**. `STATIC_DIR`/`SLOPPAK_CACHE_DIR` (which *are* patched via `setattr`) stay in `server.py` until a router that reads them is extracted, and get retargeted then.
2. **`_clean_str` → `lib/reqfields.py`.** The pure request-field sanitizer (14 callers across playlists/collections/profile/progression/shop/stats). `server.py` imports it back. Unblocks `wanted`/`saved`/`collections`/… later.
3. **`routers/playlists.py`.** Bodies verbatim; `@app`→`@router`, `meta_db`→`appstate.meta_db`, `CONFIG_DIR`→`appstate.config_dir`, `_clean_str` from `reqfields`, `_ART_CACHE_HEADERS` as a local const (art keeps `server.py`'s). The two exclusive cover helpers move with it.

## Verification
- **Route table identical** to `origin/main` (143, paths/methods/order).
- **One test retarget**: `test_playlists_api` called `server._playlist_cover_path` directly → now imports it from `routers.playlists` (which reads `appstate.config_dir`, configured by the `server` fixture). This is the seam working: the helper resolves the fresh config dir across a pop-and-reimport without the test knowing appstate exists.
- **Boot smoke** drives create / rename / add-song / cover-upload / serve / delete: the cover writes `1.png` under `CONFIG_DIR` **through `appstate.config_dir`** and serves `200` with an mtime cache-bust token; a wrong-typed `name` field still `400`s via `_clean_str`.
- `pytest` **2401 passed** (28 in playlists+collections+appstate); packaging guard **51** (auto-picked up `reqfields`); `pyflakes` clean; `npm run lint` 0; **Codex 0 findings**.

Next: `wanted`/`saved` now unblock (`_clean_str` homed). The path-`setattr` routers (`ws_highway` @ 902 lines, `song`, `audio-local-path`) need `STATIC_DIR`/`SLOPPAK_CACHE_DIR` in the seam + their 8 test retargets; `library`/`collections` need the provider registry; `enrichment` needs the network cluster. Those are the remaining substrates.

## Summary by CodeRabbit

* **New Features**
  * Added playlist management, including creation, editing, deletion, song management, and full reordering.
  * Added custom playlist cover uploads, retrieval, replacement, and removal.
  * Playlist listings now include cover images with automatic cache refresh.
* **Bug Fixes**
  * Improved handling of missing, trimmed, or incorrectly typed request fields.
  * Added validation for playlist names, types, song data, and reorder requests.
  * System playlists remain protected from renaming and deletion.
* **Documentation**
  * Updated changelog and size-exemption documentation to reflect current router extraction progress.


byrongamatos:
Both cover findings are real and both are **pre-existing** — the flagged lines (`tmp = cover.with_suffix(".png.tmp")`, `except Exception as e: … status_code=400`, the `get_playlist(pid)` existence check) are byte-identical to `origin/main:server.py`. This PR moved the cover handlers verbatim, so fixing them here would break the move-only contract.

Handling them in a follow-up, same as the loops races on #839 → #840:

- **Minor / info-leak** (400 vs 500, leaking `e`): worth fixing cleanly — split PIL decode/validation (`400`) from `img.save`/`tmp.replace` failures (log + generic `500`, no filesystem detail). Small and safe.
- **Major / cover lifecycle race + shared `.tmp`**: I'll do the cheap, substantive parts — a unique same-dir temp filename and an existence re-check immediately before the atomic replace. A full per-playlist critical section is overkill here: FeedBack is single-user (Principle I), so a cover upload racing a playlist delete on the same id essentially can't happen. I'll document that rationale rather than add a locking framework for a race the deployment model precludes.

Follow-up PR incoming; will link it and resolve these threads.

byrongamatos:
Follow-up fix opened: https://github.com/got-feedBack/feedBack/pull/842 — splits cover decode (400) from persist (500, logged, no leak), unique temp file + existence re-check. Both findings addressed; negative-checked 3 ways.

PR #844, opened 2026-07-10
refactor(server): extract the highway WebSocket into routers/ws_highway.py (R3)

byrongamatos:
The **single largest handler in `server.py`** — the 902-line `/ws/highway/{filename}` chart streamer — plus its 3 exclusive helpers. **`server.py`: 9,008 → 8,003 (−1,005, the biggest single R3 cut).**

## Why it's a clean move despite the size

The handler's only server-module dependencies are: the 4 path constants, 3 exclusive helpers (`_pick_smart_arrangement`, `_sanitize_authors`, `_sanitized_song_offset`), `app`, and `log`. Everything else it touches is either **nested inside the handler** (`_evict_audio_cache`, `_fill_scale_degree`, `_manifest_entries`, `_tone_names`, `_xml_rank`, `_send_keepalives`) or **imported from the shared lib modules** (`song`, `audio`, `sloppak`, `drums`, `notation`, `dlc_paths`, `metadata_db`). No shared server.py function is called.

## What changed

- **Path constants via the seam**: added `static_dir` / `sloppak_cache_dir` / `audio_cache_dir` slots (`config_dir` already there); `server.py` configures them. Otherwise verbatim — `@app.websocket` → `@router.websocket`, `PATHS` → `appstate.*`, `log` → module logger.
- **`sloppak_cache_dir` is `setattr`-patched**, so the 3 `test_highway_ws_*` suites now also `setattr(appstate, "sloppak_cache_dir", …)` next to their existing `server` patch. The `_sanitize_authors` unit tests import it from `routers.ws_highway` (it moved). No other test churn.
- **Removed 23 now-dead imports** from `server.py` — diffed against the `origin/main` unused-import baseline so only *newly*-dead ones went (`song`/`audio`/`drums`/`notation`/`bisect`/`contextvars`/`structlog`/`WebSocket*`/`_arr_smart_sort_key`).
- `owns_tmp` (assigned, never read) moved verbatim — pre-existing dead on `origin/main` too; left as-is to keep the move faithful.

## Verification

- **Route table identical** to `origin/main` (143, paths/methods/order); handler body verbatim spot-checked.
- `pyflakes` clean (server has no new undefined/dead); `pytest` **2400 passed**; packaging guard **53**; `npm run lint` 0.
- **WS boot smoke**: the highway WebSocket streams the full chart (`song_info` / `beats` / `sections` / `notes` / `chords` / `notation` / `anchors` / `drum_tab` → `ready`) with the **same message sequence as `origin/main`** across arrangements 0/1/2, zero tracebacks. (Confirmed against a side-by-side `origin/main` worktree.)
- **Codex**: 0 findings.

## Summary by CodeRabbit

- **New Features**
  - Added the Highway WebSocket experience for streaming song data, including notes, chords, beats, sections, lyrics, notation, tempo, and tone changes.
  - Added support for loading songs from loose folders, archives, and sloppak bundles.
  - Improved arrangement selection using explicit choices, instrument routing, preferences, and intelligent fallback.
  - Added playable audio conversion and caching, including stem and original-audio support where available.
  - Added song metadata such as credits, tuning, capo, timing, and arrangement details.


byrongamatos:
Real and **pre-existing**: the inner `except WebSocketDisconnect: pass` only wraps the post-`ready` keep-alive `receive_text()` loop. A client that disconnects *during* streaming (before `ready`) raises `WebSocketDisconnect`, which isn't caught there and falls through to the outer `except Exception as e:` → `log.exception("highway_ws unhandled error")`. So a routine tab-close mid-load logs an error. The structure is byte-identical to `origin/main:server.py`, so this move correctly didn't change it.

Follow-up incoming (same pattern as the loops/playlists follow-ups): add `except WebSocketDisconnect: return` before the blanket handler, matching the two localized guards already in the streaming body and the lib/ coding guideline. Will link it and resolve this thread.

byrongamatos:
Follow-up fix opened: https://github.com/got-feedBack/feedBack/pull/845 — adds `except WebSocketDisconnect: return` to the outer handler, with a negative-checked regression test.

PR #849, opened 2026-07-10
refactor(highway): lift 79 per-instance closure vars into hwState (R3c H lift)

byrongamatos:
The **`H`-container lift** — the enabler for splitting `highway.js`. Collapses `createHighway()`'s 79 mutable closure `let`s into one per-instance `hwState` object, so extracted renderer/ws modules can later close over `hwState` as a factory argument without multi-panel plugins (`highway_3d`, `note_detect`, splitscreen) sharing one highway's state.

Depends conceptually on the frame-time gate in #848 (the `--song` harness mode); functionally independent.

## The transform

Scope-resolved rewrite via **acorn + eslint-scope** (not regex — that would corrupt the 4 shadowed names): 1059 edits, 1057 references + 79 declaration ids − the deleted `let canvas, ctx, ws`. The four names shadowed in inner scopes (`chartTime`, `ctx`, `notes`, `chordTemplates`) are resolved correctly, so only closure-bound references move. Verified: re-parses, 0 leftover mutable closure vars, and the added lines are **all** `hwState.` (plus 2 comment lines).

Container is **`hwState`, not `H`** — `H` is already the canvas-height local (70 uses). The frame gate caught the collision instantly (0 draws, `H._drawHooks is not iterable` in the shared draw-hook path).

## Perf — the whole risk — holds

Frame-time gate (`docs/perf-baseline.md`), Arcturus feedpak, headless:

| | draw p50 | draw p95 |
|---|---|---|
| pre-lift | ~2.2 ms | 2.7–3.2 ms |
| **post-lift** | **2.1–2.2 ms** | **2.7–3.0 ms** |

Each closure-slot read became a `hwState.<slot>` monomorphic property load on a stable-shape object; the 60fps loop pays nothing. **On-device: Byron confirmed the 2D highway plays smoothly.**

## Tests

The ~30 highway JS suites brace-extract functions/patterns from the source; their state references and the `monotonic_clock` vm sandbox now use `hwState.<slot>` (consts like `_CHART_MAX_INTERP_MS` stay top-level — not lifted). **`node --test`: 1030/1030 green.** Two self-inflicted over-replacements caught and reverted along the way (`_lefty` is a *prefix* of the 3D-local `_leftyCached`; `STRING_COLORS` a *suffix* of `DEFAULT_STRING_COLORS`) — substring replaces on brace-extract regexes need word care.

`npm run lint` 0 errors; **Codex 0 findings**. Transformer saved for reproducibility.

Next (R3c cont.): with `hwState` in place, the renderer/ws/geometry modules can be extracted as `createRenderer2d(hwState)` etc.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved renderer stability during WebSocket reconnects and renderer mode swaps.
  - Fixed timing/pause detection and made draw-loop behavior more consistent across playback states.
  - Improved visibility throttling and ensured HUD/canvas state resets correctly.
  - Enhanced adaptive render scaling behavior while playing.
- **Tests**
  - Updated assertions to match the new instance-scoped renderer state across timing, visibility, note/chord/handshape rendering, renderer swapping, and adaptive scaling.

### DIFF
--- a/static/highway.js
+++ b/static/highway.js
@@ -3,24 +3,27 @@
  * Receives note data via WebSocket, renders on requestAnimationFrame.
  */
 function createHighway() {
-    let canvas, ctx, ws;
+  // R3c: per-instance mutable state in one object, so extracted renderer/ws
+  // modules can close over it as a factory arg without cross-panel sharing.
+  const hwState = {};
+    
     // Promise chain for serializing async ws.onmessage handlers —
     // reset on each connect() so reconnections start a fresh chain.
-    let _msgChain = Promise.resolve();
+    hwState._msgChain = Promise.resolve();
     // Monotonically-increasing connection generation counter.
     // Incremented on every connect() so async handlers that survive a
     // reconnect can detect they are stale and bail out before mutating
     // shared state.
-    let _wsGen = 0;
+    hwState._wsGen = 0;
     // The audio-element accessor below is polled by plugins, so record its
     // legacy bridge hit only once per highway instead of on every call —
     // otherwise a per-frame poller floods playback:bridge-hit events and
     // diagnostics.
-    let _audioElementBridgeRecorded = false;
+    hwState._audioElementBridgeRecorded = false;
     // Pending JUCE routing promise for the current connection's song_info.
     // The 'ready' handler awaits this so _juceMode is settled before
     // _onReady / song:ready fire, without blocking note/chord processing.
-    let _juceRoutingPromise = Promise.resolve();
+    hwState._juceRoutingPromise = Promise.resolve();
 
     function _audioSession() {
         return window.feedBack && window.feedBack.audioSession;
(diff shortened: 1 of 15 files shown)


byrongamatos:
On the two threads:

**max-lines (4224 vs 1500)** — expected and tracked. `highway.js` is the planned-split monolith; it sits in `docs/size-exemptions.md` under *Planned, NOT exempt* with a pre-approved post-split `renderer-2d.js` ceiling of 3,000. This PR is the **state-lift prerequisite** (the `hwState` container is what lets the renderer/ws/geometry modules become `createRenderer2d(hwState)` factories); the module split that brings the residual under the ceiling is the next R3c step. `max-lines` is a warning, not an error — CI lint is green. The net line delta here is ~+3 (895 ins / 892 del).

**note_state:42** — fixed in `<pushed>`: the assertion now requires `hwState._noteStateProvider =`, so a regression to closure/shared state fails the test instead of matching an unqualified form.

PR #876, opened 2026-07-11
refactor(ui): load app.js as an ES module (R3a) — the flip

byrongamatos:
**The flip.** One attribute. #871, #872, #874 and #875 exist to make this line safe.

```diff
-<script defer src="/static/app.js"></script>
+<script type="module" src="/static/app.js"></script>
```

After this, `app.js` can `import` — the monolith-killing *enabler* is done, and carving it into `static/js/*.js` becomes routine follow-on work.

## What changes

app.js's **385 top-level `function` declarations** stop being implicit `window` properties:
- **87** stay reachable through the explicit contract — #874's `Object.assign` block **plus 47 pre-existing scattered `window.X = X` assignments**
- **298** become module-private

Verified: **no unexposed name is read from outside app.js.**

## Strict mode

Modules are always strict, and app.js has never run under it. Checked before flipping:

- parses clean as `sourceType: "module"` (no octal literals, duplicate params, `with`)
- no implicit globals, no `eval` / `new Function`, no top-level `this`
- **`registerShortcut`** is called **bare** at 15 top-level sites. It's assigned at `window.registerShortcut` (`app.js:10387`), *before* its first call (`10648`), and a bare identifier in a module still resolves through the global object. Confirmed `typeof window.registerShortcut === 'function'` in the browser.

## Hard gate — app.js *is* the plugin loader

| check | result |
|---|---|
| `/api/plugins` `script_type` passthrough | editor / stems / studio = `"module"` |
| `/api/plugins/stems/src/main.js` | **200** |
| conditional GET (live-edit ETag) | **304** |
| deep graph: `stems/src/transport.js`, `editor/src/state.js` | **200** |
| `window.loadPlugins` | `function` |
| plugin screens mounted | 5 |
| migrated plugins injected as `<script type="module">` | editor, studio, stems |
| capability participants / compat shims | 37 / 14 |

**All 336 inline handlers in the shell resolve on `window`** under module scope; the A–Z rail and pagination execute **6/6** with no ReferenceError.

A/B against `origin/main`: the only unresolvable handler is **`editorToggleStemMixer`** — **equally broken on main**. It's a dead handler in the editor plugin (not defined anywhere in its source — looks like rename skew). Pre-existing, unrelated to this PR, worth its own issue.

## Codex preflight — one [P1], a false positive

Codex claimed `restartCurrentSong`, `requestExitSong`, `editRegionInEditor` and `returnToEditorFromHighway` would `ReferenceError` because they're absent from #874's `Object.assign` block.

They are absent from it — because they were **already** explicit, among app.js's 47 scattered `window.X = X` assignments:

```
app.js:7086  window.restartCurrentSong = restartCurrentSong;
app.js:7204  window.requestExitSong = requestExitSong;
app.js:8492  window.editRegionInEditor = editRegionInEditor;
app.js:8511  window.returnToEditorFromHighway = returnToEditorFromHighway;
```

With app.js loaded as `type="module"`, all four resolve as `function` in the browser, and `requestExitSong()` is live as a real `onclick` source. The contract test already covers this — it collects **both** the scattered assignments and the block. Codex's scan only read the block.

## Suites

pytest **2396 passed** · node **1032/1032** · ESLint **0 errors** · tailwind-fresh clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application startup behavior by updating how the app’s main script loads in the browser.
  * Enables modern module-based loading for better compatibility and execution handling.

### DIFF
--- a/static/v3/index.html
+++ b/static/v3/index.html
@@ -1244,7 +1244,7 @@
     <script defer src="/static/highway.js"></script>
     <script defer src="/static/vendor/lottie.min.js"></script>
     <script defer src="/static/lottie-api.js"></script>
-    <script defer src="/static/app.js"></script>
+    <script type="module" src="/static/app.js"></script>
     <script defer src="/static/audio-mixer.js"></script>
     <script defer src="/static/vendor/shepherd.min.js"></script>
     <script defer src="/static/tour-engine.js"></script>


byrongamatos:
@coderabbitai Thanks — but this is a **false positive**, and interestingly it's the exact mirror of the one Codex raised on this same PR.

app.js's window surface is **split across two forms**:

| form | count | example |
|---|---|---|
| scattered `window.X = X;` statements | **47** | `window.restartCurrentSong = restartCurrentSong;` (app.js:7086) |
| the `Object.assign(window, { … })` contract block added in #874 | **66** | `showScreen, togglePlay, setSpeed, …` |

Your script collected only the first form:

```
rg -oP 'window\.\K[A-Za-z_$][\w$]*' static/app.js    # ← 47 names; never parses the Object.assign block
```

…so all 66 block names look "missing". **Every one of the 13 you listed is in the block:**

```
showScreen           Object.assign block
uploadSongs          Object.assign block
setLibraryProvider   Object.assign block
setLibView           Object.assign block
sortLibrary          Object.assign block
toggleLibFilters     Object.assign block
setFavView           Object.assign block
setAvOffsetMs        Object.assign block
persistSetting       Object.assign block
setLoopStart         Object.assign block
changeArrangement    Object.assign block
setSpeed             Object.assign block
togglePlay           Object.assign block
```

And at runtime, with `app.js` actually loaded as `<script type="module">`:

```
app.js is type=module: True
showScreen: function | uploadSongs: function | setLibraryProvider: function | setLibView: function | sortLibrary: function
toggleLibFilters: function | setFavView: function | setAvOffsetMs: function | persistSetting: function | setLoopStart: function
changeArrangement: function | setSpeed: function | togglePlay: function

UNDEFINED: NONE — all 13 resolve
```

More broadly: I drove **all 336 inline handlers** in the shell under module scope and resolved each one's leading identifier against `window`. The only unresolvable name is `editorToggleStemMixer` — which is **equally broken on `main`** (the editor plugin's `screen.html:42`/`:177` call it, but the plugin defines `editorToggleMixer`; pre-existing rename skew, its own PR).

Exposing these 66 names is precisely what #874 did, as a standalone no-op PR, ahead of this flip. `tests/js/window_contract.test.js` guards it and reads **both** forms — which is why it passes. It's also verified to *bite*: removing `showScreen`, `goTreePage`, or `setViz` each fails it.

**No change needed.** That said, two independent reviewers misread the same split surface in opposite directions, which is a fair signal that consolidating the 47 scattered assignments into the single block is worth a follow-up cleanup — just not inside a move-only PR.

Issue #98, opened 2026-06-20
Surface arrangement dropdown for direct selection inside Tab View (today the user picks...


byrongamatos:
> _Restored from `slopsmith/slopsmith#589` — original issue, opened by @byrongamatos on 2026-05-09._
> _[restored-from: slopsmith/slopsmith#589]_

_From `specs/001-tab-view/tasks.md` — generated retrospectively by spec-kit._

Surface arrangement dropdown for direct selection inside Tab View (today the user picks it from the player UI before switching).

---
See `specs/001-tab-view/spec.md` and `specs/001-tab-view/analyze.md` for context.
This issue was filed in bulk; close as `wontfix` if no longer relevant.

PR #877, opened 2026-07-11
feat(audio): loopback feeder mode + static no-cache — all app audio under exclusive/ASIO

OmikronApex:
Builds on the merged renderer-bus work (#852/#865). Companion desktop PR: got-feedBack/feedBack-desktop#98.

## Loopback feeder mode (new preferred)
One `getDisplayMedia` frame-audio capture covers **every** sound the app makes — song, previews, UI — for the whole exclusive/ASIO session (engages even with no song loaded; fixes tester report of previews/UI on the default WASAPI device). Local playback silenced via `suppressLocalAudioPlayback` (spike-verified), page-mute IPC fallback otherwise. Sticky fallback to the existing stems/element surface modes when capture is unavailable; Docker sphere untouched.

## Element-capture poisoning fix (tester log, confirmed)
`createMediaElementSource(#audio)` collides with highway_3d's one-shot analyser tap; `_elCtx` was assigned pre-throw → every later tick died in a TypeError while the song played on the default device. Module state now assigned only on full success; failed engage disables the bus and tears down loopback.

## /static Cache-Control: no-cache
Root cause of the fix14 tester regression: without Cache-Control, Chromium's heuristic freshness served a previous build's `app.js` for days — the routing watcher/feeder never ran and the (old) plugin loader broke module plugins. `no-cache` keeps caching but forces ETag revalidation (unchanged files = 304). Benefits the Docker sphere equally.

## Diagnostics (gated on desktop --debug via `audio.debugEnabled()`)
Install/NOT-installed tracing for watcher + feeder, global uncaught-error tap with file:line:col, full device object on outputType change, feeder decision vector, sink flips, DOMException details. Docker sphere: fully silent.

## Verification
31/31 JS tests (12 feeder — 5 new for loopback + collision retry — and 19 reroute); `node --check` clean; fix15 portable build smoke-tested cold-launch end-to-end.

## Summary by CodeRabbit

* **New Features**
  * Added whole-app “loopback” audio capture for exclusive-output sessions.
  * Automatically falls back to element or stem capture when loopback can’t be used.
  * Added page playback muting/unmuting when local playback suppression isn’t supported.
* **Bug Fixes**
  * Improved capture recovery after setup failures and prevented silent/stranded audio states after failed engagement attempts.
  * Better handling for cases where audio elements are already in use by another capture path.
* **Chores**
  * Updated `/static` responses to default to no-cache behavior.


OmikronApex:
Companion desktop PR: got-feedBack/feedBack-desktop#98

PR #878, opened 2026-07-11
refactor(app): carve the plugin loader out of app.js into static/js/ (R3a)

byrongamatos:
The **first carve**, and deliberately the riskiest one: **app.js *is* the plugin loader** (the R0 host rails). It goes first, while the module graph is still one edge deep and a mistake is easy to see.

`static/js/plugin-loader.js` (829 lines) — bodies **verbatim**. `app.js` **12,217 → 11,439**. Core's first `static/js/` module, which is exactly the shape constitution §II anticipates.

## The closure

Measured with **acorn**, not regex — brace-matching the comment-stripped source drifted, and this is precisely the kind of file where an off-by-one is a silent disaster.

The block at `app.js:11246-12031` turns out to be contiguous and self-contained. It needs **two** things from the rest of app.js and exports **two**:

| direction | symbol | handling |
|---|---|---|
| export | `loadPlugins` | the `window` contract |
| export | `bootstrapPluginsAndUi` | called at boot |
| inbound | `window.showScreen` | **already** the public host contract (§II) — called through `window`, not re-coupled as an import |
| inbound | `_populateVizPicker` | injected via `configurePluginLoader()` |

## Why a seam and not an import

`plugin-loader` must not import `app.js` — app.js imports *it*, so that closes a cycle.

I checked whether `_populateVizPicker` could simply **move into** the module, which would delete the seam entirely. It drags **9 further symbols** (`_canRun3D`, `_autoMatchViz`, `_showPromotionNag`, `_ensureVenueVizOption`, …) — a whole viz cluster. That's its own carve. So the seam stays.

## The seam's default is loud, on purpose

A silent no-op stub is the classic failure for this pattern — the editor hit it twice. Drop the wiring call and the loader keeps working while the viz picker quietly stops refreshing: no test, no boot check, nothing says a word.

So the default hook `console.error`s instead, which the smoke harness (it treats console errors as failures) will catch.

**Bite-tested.** Removing `configurePluginLoader()` from app.js surfaces:

```
[plugin-loader] host seam not configured — app.js must call configurePluginLoader({ populateVizPicker })
```

That also proves the seam **is** exercised on the plugin-startup path, so an unwired hook cannot pass silently. Without this check, my A/B would have been blind to it: the viz picker is empty in the smoke env either way.

## `no-cycle` is live on core's own graph for the first time

`eslint.config.js` gains `static/app.js` + `static/js/**` to the module block — app.js now `import`s, so parsing it as a script would be a syntax error.

**Bite-tested:** making `plugin-loader` import `app.js` back fails with `Dependency cycle detected`.

## Harnesses — the R3a note said budget one conversion per carve. It was five.

Retargeted: `capability_inspector_nav`, `plugin_hydration_wipe`, `plugin_loader_script_type`, `plugin_style_injection`, `legacy_shim_hits`, plus `test_plugin_runtime_idempotence.py`.

Two things worth knowing for the next carve:

- **`legacy_shim_hits` was missed by a symbol-name grep** because it greps for a code *string* (``script.src = `/api/plugins/${plugin.id}/screen.js` ``), not a symbol. Only the failing run found it. It's also **split** — one test needs the loader, one still needs app.js.
- **`test_capability_events`' assertions are NEGATIVE** (`assert token not in source`). Carving code out of app.js would have made them **vacuous instead of failing**. They now span app.js **+** the loader.

## Verification

**A/B against `origin/main`** in two browsers — mounted plugin screens, 14 loaded plugin scripts, the 3 module plugins injected as `<script type="module">`, 37 capability participants, 14 compat shims, `window.loadPlugins`: **IDENTICAL, zero console/page errors on both.**

`/static/js/plugin-loader.js` serves **200**. R0 rails intact: `stems/src/main.js` → 200, conditional GET → **304**, `script_type` passthrough.

pytest **2396** · node **1032/1032** · ESLint **0 errors** · tailwind-fresh clean · Codex **0**.

## Codex caught a real P1 on the first pass

`static/js/plugin-loader.js` was **untracked** — I'd never `git add`ed it. A checkout of the patch would have served an `app.js` importing a module that doesn't exist, and **a failed static import kills the entire module** and every `window` handler with it. Now tracked. Worth flagging because a green local run says nothing about this.

## Summary by CodeRabbit

* **New Features**
  * Improved plugin loading and UI updates, including support for repeated refreshes during installation and upgrades.
  * Added reliable handling for plugin scripts, styles, navigation, settings, and visualization integrations.
  * Added startup status monitoring with streaming updates and polling fallback.

* **Refactor**
  * Plugin management is now handled through a dedicated module, improving maintainability without changing expected behavior.

* **Tests**
  * Updated plugin-related tests to validate the new module structure and loading behavior.


byrongamatos:
@coderabbitai Good catch, and the analysis is correct — but it's **pre-existing, not introduced here**, so I'm deferring it rather than riding it along on a move-only PR.

**It's a verbatim move.** The URL construction and module injection are byte-identical to `origin/main`:

| | line | code |
|---|---|---|
| `origin/main` `static/app.js` | 11884 | ``script.src = `/api/plugins/${plugin.id}/screen.js${v ? `?v=${v}` : ''}`;`` |
| this branch `static/js/plugin-loader.js` | 682 | ``script.src = `/api/plugins/${plugin.id}/screen.js${v ? `?v=${v}` : ''}`;`` |

(same for `script.type = 'module'`, main:11892 → loader:690). It arrived with the R0 module rails in #812; this PR only changed which file it lives in.

**The bug is real though.** `?v=<plugin.version>` is keyed on the version, not on the load, so a **downgrade** — rolling a plugin back to a version already evaluated this session — reuses a URL that's already in the module map. Modules evaluate once per URL per document, so the script silently doesn't re-run while `loadedScripts` records that it did. The screen stays stale. Upgrades are fine (new version → new URL); it's specifically the rollback path.

Deferring because the fix is a **behaviour change to the plugin-loading path** (a monotonic load token changes the URL every reload, which also defeats the ETag/304 live-edit caching the R0 rails deliberately rely on — that interaction needs thinking about, not a one-liner). This PR is a verbatim carve whose entire value is being provably behaviour-neutral; folding a real fix into it would destroy that property and give the on-device pass nothing to compare against.

Filing it as its own issue with your analysis. Thanks — this is a genuinely non-obvious one.

PR #912, opened 2026-07-12
refactor(highway): make the `highway` global explicit before the module flip (R3c)

byrongamatos:
73 bare `highway.x` references → `window.highway.x`, across app.js and 10 other files. **Provably a no-op today.** It is the precondition for flipping highway.js to a module.

## Why this has to land first

highway.js is a **classic script**. Its top-level `const highway = createHighway()` therefore creates a **global lexical binding** — visible as a bare name to every other classic script *and* to every ES module. 73 call sites quietly rely on that.

**The moment highway.js becomes a module, that binding is gone.** `const` in a module is module-scoped, not global. Every one of those 73 sites becomes a `ReferenceError`, and the flip is impossible until they say what they mean.

`window.highway = highway` is already set, to the same object, on the same line — so this is an identity rewrite. Verified in the browser below.

## The rewrite bit me three times. Regex is not enough for this.

**1. A shadowed local.** `capabilities/note-detection.js` does `const highway = window.highway`. Its 9 bare uses are *local* and already correct; a blind rewrite would have emitted `const window.highway = window.highway`. Excluded.

**2. Half-converted guards — the dangerous one.** Six sites read:

```js
typeof highway !== 'undefined' && highway && typeof highway.setTime === 'function'
```

The regex converted the **consequent** and left the **test** — which is *worse* than not touching them. After the flip, `typeof highway` is `'undefined'`, so each guard is **permanently false** and the code behind it silently never runs.

transport.js's was the **seek → `setTime` sync**: the chart clock would have quietly desynced after every seek, with nothing failing. All six now test `window.highway`.

**3. Two more bare references** — found by **Codex [P2]** and confirmed by an AST scan. `app.js:3114` and `:3176` use `highway && typeof window.highway.getSections === 'function'`. My grep searched for `typeof highway`, not `highway &&`. After the flip these throw, the `catch` swallows it, and the editor silently falls back to a ±4s edit window and arrangement 0.

Regex missed **a shadow, a half-conversion, and two bare reads.** The final check is an AST pass that resolves scopes and reports every `highway` identifier not bound locally. It now reports **zero**.

## Verification

A/B against `origin/main` in two browsers — **15 probes, identical, zero page errors**:

| | |
|---|---|
| `window.highway` **is** the bare global (same object) | ✓ |
| the whole API surface resolves (`getTime`, `setTime`, `getBPM`, `getSections`, `setAvOffset`, `getSongInfo`, `getPerf`) | ✓ |
| a real song plays; the chart clock advances | ✓ |
| `getPerf().drawMs > 0` | ✓ |
| **`seek syncs chart`** — the exact guard I nearly broke in (2) | ✓ |

node **1045** · pytest **2416** · ESLint **0** · Codex **0**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved player lifecycle handling when leaving and restarting playback.
  * Preserved resume-session snapshots more reliably during player transitions.
  * Improved synchronization for playback time, looping, lyrics visibility, mastery controls, and section practice.
  * Prevented errors when certain playback services are unavailable.
* **Tests**
  * Updated playback tests to reflect improved service initialization and synchronization behavior.

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -860,10 +860,10 @@ async function showScreen(id) {
         const hadPlayableSong = !!audio.src || !!window._juceAudioUrl || S.isPlaying;
         // Snapshot where we were so leaving the player — especially by accident
         // — is recoverable instead of dumping the user back at bar 1 next time.
-        // Must run BEFORE highway.stop()/audio unload, while getSongInfo() and
+        // Must run BEFORE window.highway.stop()/audio unload, while getSongInfo() and
         // the position (stopTime) are still live.
         if (hadPlayableSong) _snapshotResumeSession(stopTime);
-        highway.stop();
+        window.highway.stop();
         // Cancel any queued seeks, in-flight shim closures, AND active
         // count-in timers before stopping playback so none of these paths
         // can mutate the torn-down session (mirrors the same triple reset
@@ -1049,7 +1049,7 @@ async function loadSettings() {
     const demucsEl = document.getElementById('demucs-server-url');
     if (demucsEl) demucsEl.value = data.demucs_server_url || '';
     const leftyEl = document.getElementById('setting-lefty');
-    if (leftyEl) leftyEl.checked = highway.getLefty();
+    if (leftyEl) leftyEl.checked = window.highway.getLefty();
     const autoplayExitEl = document.getElementById('setting-autoplay-exit');
     if (autoplayExitEl) autoplayExitEl.checked = _autoplayExitEnabled();
     const showUpNextEl = document.getElementById('setting-show-upnext');
@@ -1435,7 +1435,7 @@ function setAvOffsetMs(ms, skipPersist) {
     // the audio-aligned chart time so plugins (note detection, etc.)
     // keep scoring against the real chart clock regardless of visual
     // calibration.
-    if (typeof highway !== 'undefined' && highway?.setAvOffset) highway.setAvOffset(_avOffsetMs);
+    if (window.highway?.setAvOffset) window.highway.setAvOffset(_avOffsetMs);
     // Sync any visible Settings slider
     const avSlider = document.getElementById('setting-av-offset');
     if (avSlider) {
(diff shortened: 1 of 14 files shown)


byrongamatos:
## Correction: this PR's stated premise was wrong

Posting this against the merged PR so the commit message doesn't mislead whoever reads it next.

I claimed:

> The moment highway.js becomes a module, that binding is gone. […] Every one of those 73 sites becomes a `ReferenceError`.

**That is false.** I have now actually flipped highway.js to a module and measured it:

```
main (classic script)        flipped (module)
  bare `highway` resolves       bare `highway` resolves
  window.hasOwnProperty: true   window.hasOwnProperty: true
```

`highway.js` already did `window.highway = highway` — which puts the name on the **global object**, and bare-identifier resolution falls back to the global object regardless of whether a *lexical* global binding exists. So the 73 sites would have kept working. The flip does not break them.

### What that means for this PR

The rewrite is still defensible as hygiene — `window.highway.x` says what it means, and it stops the code depending on an implicit global-object fallback. But it was **not** a precondition for the flip, and it did **not** fix a latent bug.

Worse: the "six half-converted guards" and "two bare reads" I reported as finds were **bugs my own regex introduced and then repaired**. They were not pre-existing. Net, this PR is churn plus a small hygiene gain, and I over-sold it.

### What the real hazard turned out to be

Not `window.highway` — **`window.createHighway`**.

A classic script's top-level `function createHighway()` *implicitly* becomes `window.createHighway`. A module's does not. The constitution names `window.createHighway` as public extension contract, and nothing in-tree calls it — the only consumers are third-party plugins, which I cannot grep.

Verified by removing the explicit assignment from the flip:

```
flip WITHOUT an explicit assignment:  window.createHighway === undefined   <-- contract silently gone
flip WITH it:                         window.createHighway === function
```

**That** is the thing the flip had to preserve, and it is handled in the flip PR. It is also precisely the class of breakage this PR's premise claimed to be preventing — I just had the wrong name.

Issue #97, opened 2026-06-20
Vibrato variants (wide / shallow distinction)


byrongamatos:
> _Restored from `slopsmith/slopsmith#590` — original issue, opened by @byrongamatos on 2026-05-09._
> _[restored-from: slopsmith/slopsmith#590]_

_From `specs/001-tab-view/tasks.md` — generated retrospectively by spec-kit._

Vibrato variants (wide / shallow distinction).

---
See `specs/001-tab-view/spec.md` and `specs/001-tab-view/analyze.md` for context.
This issue was filed in bulk; close as `wontfix` if no longer relevant.

Issue #105, opened 2026-06-20
Add random sort option


byrongamatos:
> _🗂️ Migrated from the prior repo. Originally opened by @mogul on 2026-05-14 (was #188)._

> _Restored from `slopsmith/slopsmith#268` — original pull request (open PR; branch did not survive redaction, archived as issue), opened by @Eatkin on 2026-05-14._
> _[restored-from: slopsmith/slopsmith#268]_

# Summary

Adds random as a sort option in the library, using SQLite's RANDOM() function.

# Changes

- Add `"random": "RANDOM() ASC"` to sort_map in query_page
- Add option to sort dropdown in `index.html`

# Verification

Tested via the library sort dropdown locally - songs shuffled as expected

byrongamatos:
> _💬 @mogul on 2026-06-18:_

> _[restored-pr-reviews: slopsmith/slopsmith#268]_
> _Reconstructed from the backup. The PR's conversation-thread comments were not captured; reviews and inline review comments follow._

## Reviews

**@coderabbitai[bot]** — COMMENTED — 2026-05-14

**Actionable comments posted: 1**

<details>
<summary>🤖 Prompt for all review comments with AI agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

Inline comments:
In `@server.py`:
- Line 515: The current mapping for sort "random" uses "RANDOM() ASC", which
re-shuffles on every request; change the implementation that builds the ORDER BY
for sort="random" to use a deterministic per-session seed: accept/validate a
seed query param (e.g., seed string or integer), and replace the ORDER BY
expression "RANDOM() ASC" with a stable expression derived from the seed and a
stable column like filename (for Postgres: ORDER BY md5(concat(:seed, filename))
ASC or for other DBs use equivalent hash/concat), and ensure the code path that
constructs the SQL (the place that currently maps "random" -> "RANDOM() ASC")
uses that expression and falls back to a default seed when none provided.
```

</details>

<details>
<summary>🪄 Autofix (Beta)</summary>

Fix all unresolved CodeRabbit comments on this PR:

- [ ]  Push a commit to this branch (recommended)
- [ ]  Create a new PR with the fixes

</details>

---

<details>
<summary>ℹ️ Review info</summary>

<details>
<summary>⚙️ Run configuration</summary>

**Configuration used**: Organization UI

**Review profile**: CHILL

**Plan**: Pro Plus

[...]

PR #938, opened 2026-07-13
feat(v3): add desktop-only "Start in fullscreen" system option

gionnibgud:
## What

Adds a **Start in fullscreen** toggle to Settings → System (between App updates and Library folder path).

Addresses the desktop request in got-feedback/feedBack-desktop#97 — users want the app to launch fullscreen without pressing the OS hotkey (⌃⌘F) every time.

## How it stays desktop-only

The block ships `hidden` and is gated exactly like the existing App-updates block. `setupWindowOptions()` (in `static/js/settings.js`, called from `loadSettings()`) only unhides + wires it when the feedBack-desktop bridge exposes:

```js
window.feedBackDesktop.window.{ getStartFullscreen, setStartFullscreen }
```

- Web / Docker builds have no `feedBackDesktop` bridge → the function returns early → the section never appears.
- It's a **per-method capability check**, so even on desktop it stays hidden until the bridge actually ships those methods — no half-wired toggle.

Persistence lives desktop-side because only the Electron main process can read the pref at window-creation time; core just proxies through the bridge (getter normalised via `Promise.resolve` to allow sync or async IPC).

## Scope

Core UI + bridge proxy only. The desktop bridge (`feedBackDesktop.window`) and the actual launch-in-fullscreen behaviour land in a follow-up **feedBack-desktop** PR. Until then this is inert in every shipping build.

## Testing

- Web/Docker: `#window-options-block` is present in the DOM but `display:none` (no bridge) — nothing new renders. Manually `classList.remove('hidden')` in devtools confirms the row layout.
- No new Tailwind utilities (all classes pre-existing), so no `tailwind.min.css` rebuild.

## Summary by CodeRabbit

* **New Features**
  * Added a desktop-only **Window options** section in Settings.
  * Added a **Start in fullscreen** option that reflects and saves the selected preference.
  * The option stays hidden when desktop window controls aren’t available.

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -218,6 +218,7 @@ import {
     setAvOffsetMs,
     setInstrumentPathway,
     setupAppUpdates,
+    setupWindowOptions,
     syncDefaultArrangementPin,
 } from './js/settings.js';
 import {
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -100,6 +100,7 @@ export async function loadSettings() {
     // failed fetch below still leaves the desktop updater wired up.
     // setupAppUpdates() is idempotent via _appUpdatesWired.
     setupAppUpdates();
+    setupWindowOptions();
     const resp = await fetch('/api/settings');
     const data = await resp.json();
     // Null-guard the form fields: on the v3 tabbed settings page the markup is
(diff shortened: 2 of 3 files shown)


gionnibgud:
Companion desktop PR: got-feedback/feedBack-desktop#105 — adds the `feedBackDesktop.window` bridge, persists the pref, and applies `fullscreen: true` at window creation. This core PR is the UI/settings half; #105 is the Electron half. Both are needed for the feature to function end-to-end.

gionnibgud:
Thanks @coderabbitai — evaluated all three; skipping each, with rationale:

**1. Sync getter throw (`Promise.resolve(winApi.getStartFullscreen())`, settings.js) — skip.** This mirrors the existing `update`-bridge hydration in the same file (`Promise.resolve(updateApi.getStatus()).then(...)`). `getStartFullscreen` is an `ipcRenderer.invoke` call, which returns a Promise and doesn't throw synchronously, and the capability check already guarantees it's a function. Changing only this row would make it inconsistent with the sibling bridge code for a case that can't occur here.

**2. Late hydration overwriting a user toggle — skip.** The race window is a single IPC round-trip on Settings entry, and the toggle is its own sole writer — the `change` handler already persists and live-applies the user's choice via `setStartFullscreen`, so a stale read would at most briefly mis-show the checkbox and self-heals on next entry. Same hydrate-on-entry pattern as the `update` block; not worth the added interaction-tracking state.

**3. Checkbox accessible name — valid, but out of scope here.** The finding is correct, but it's a **page-wide pattern**: every sibling toggle in Settings (`setting-lefty`, `setting-confirm-exit`, `setting-show-upnext`, …) uses the identical `fb-switch` structure with the title as a sibling of the `<label>`, so they all share this gap. Fixing only this one row would be inconsistent, and reworking the switch/label a11y pattern across the whole settings page (or app) is a design decision for a dedicated pass, not this feature PR. Flagging it for the maintainers to tackle page-wide.

Net: this PR stays consistent with the established settings patterns; no changes from this review.

PR #967, opened 2026-07-14
perf(folder_library): render only the songs on screen — 1.3M DOM nodes -> ~30 (#965)

### LINKED ISSUE #965: folder_library renders the entire library as an unvirtualized DOM tree (1.3M nodes, ~4.2 GB) on a hidden screen
Found while profiling a ~2 fps report in the venue highway (see feedBack-plugin-song-preview#7, which fixes a *separate* per-frame document walk). This is the deeper cause.

## What

On a 50,944-song library, the renderer DOM holds **1,318,810 nodes**. Nearly all of them belong to one plugin:

```
TAG    ID                      CLASS                      DESCENDANTS  CHILDREN
------------------------------------------------------------------------------
MAIN   v3-main                 flex-1 overflow-y-auto...    1,318,389        39
DIV    plugin-folder_library   screen                       1,300,907         5
DIV    fb-tree                 px-2 py-2                    1,300,854         4
DIV                            mb-1                         1,300,773         2
DIV                            ml-5 mt-0.5 space-y-0        1,300,763    50,938
```

That last div has **50,938 children — one per song** — and ~1.3M descendants. The library tree is rendered in full, unvirtualized. For comparison, the v3 Songs grid *is* virtualized (58 cards in the DOM).

The visible screen at the time was `v3-home`. `plugin-folder_library` was **not even on screen**.

## Why it matters

- **~4.2 GB renderer RSS** (a fresh instance is ~400 MB).
- It makes *any* document-wide query catastrophic. A single `document.querySelector` miss must walk all 1.3M nodes — that is exactly how song_preview's per-frame menu check ate 47.7% of the renderer and dropped the app to 2.7 fps.
- Even with that fixed, the app only reaches **11.3 fps** — the remaining cost is style/layout/memory pressure from the tree itself.

So this is a landmine: it degrades unrelated code that has no idea the DOM is a million nodes deep, and it scales with library size — invisible in dev, brutal on a real collection.

## Suggested direction

Virtualize the tree (render only expanded/visible rows), or build it lazily when the screen is first shown and tear it down when it is hidden. Rendering 50k rows for a screen the user is not looking at buys nothing.

## Repro

Any library of this size. `document.getElementById("plugin-folder_library").getElementsByTagName("*").length` in the renderer console.

byrongamatos:
Closes #965.

## The bug

A song list rendered **every** song it held. On a flat 50,944-song library that is one `<div>` with **50,938 children and ~1,300,000 DOM nodes** — **~4.2 GB** of renderer RSS, for a screen the user may not even be looking at (it was built while the visible screen was `v3-home`).

```
DIV  plugin-folder_library                1,300,907 descendants
DIV  fb-tree                              1,300,854
DIV  (ml-5 mt-0.5 space-y-0)              1,300,763   <- 50,938 children
```

It is not just this plugin's problem. **A million-node document poisons unrelated code:** any `document.querySelector` that *misses* must walk the entire tree before returning null. That is precisely how song_preview's per-frame menu check ended up eating ~50% of the renderer and dropping the app to **2.7 fps**. song-preview#7 fixed the per-frame walk; this fixes the tree it was walking.

## The fix

Render only what is on screen. Rows are uniform height (grid cards uniform size), so the window is pure arithmetic — no per-row observers.

Off-window songs are represented by **padding on the list**, not spacer elements: a spacer `<div>` would become a grid **item** in grid view and shift the columns, whereas padding behaves identically in both layouts.

Lists at or below `VIRTUAL_MIN` (200) render in full exactly as before — normal folders are untouched.

**Two ordering bugs this forced out, both landmines:**
- Both expand handlers populated the list **before** showing it. A windowed list measures a real row and the scroller viewport, and both are zero under `display:none`. Show first, then populate.
- `_render()` now tears down the previous render's scroll listeners — without it they survive against detached nodes and leak on every re-render.

## Verified in real Chromium (CDP), 50,000 rows

Not just the maths — the DOM glue:

| | rendered rows | scrollHeight | showing |
|---|---|---|---|
| at top | **25** | 2,200,000px | [0..24] |
| scroll 500k | **31** | 2,200,000px | [11357..11387] |
| scroll 1,100k | **31** | 2,200,000px | [24994..25024] |
| scroll to end | **25** | 2,200,000px | [49975..49999] |

**25–31 rows in the DOM instead of 50,000.** Scroll height exact and constant (the scrollbar stays honest), and the last row lands on song 49,999.

## Tests

`_visibleWindow` is pure and exposed via `__test`: top / middle / bottom / scrolled-past-the-end windows, the grid row-packing case, the **padding + rendered = total** invariant that stops the list changing height as you scroll, and the degenerate zero-height case (a list still `display:none`) falling back to render-everything rather than to an empty list.

eslint clean; full JS suite **1186/1186**.

## Not verified

I could not drive the folder-library screen end-to-end against a real 50k library in this session (the plugin needs the library view active). The windowing is validated in the real engine with 50k synthetic rows, and small lists take the untouched code path — but a visual pass on the actual screen is worth doing before this ships to testers.

## Summary by CodeRabbit

* **Performance**
  * Improved folder library performance by rendering only songs currently visible on screen.
  * Reduced initial loading and scrolling overhead for large song collections in both grid and list views.

* **Bug Fixes**
  * Improved folder expansion and list sizing behavior.
  * Prevented stale scrolling behavior after refreshing or closing the folder library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   carry their gig log; instruments their gig count.
 
 ### Changed
+- **Folder library renders only the songs on screen** (#965) — a song list used to
+  render *every* song it held. On a flat 50,944-song library that was one `<div>`
+  with 50,938 children and ~1.3 **million** DOM nodes (~4.2 GB of renderer memory),
+  built even while another screen was showing. A document that size also punishes
+  unrelated code: any `document.querySelector` that misses has to walk the whole
+  tree — which is how the song-preview menu check ended up eating ~50% of the
+  renderer and dropping the app to 2.7 fps. Lists longer than 200 songs are now
+  windowed (25–31 rows in the DOM instead of 50,000); shorter lists are unchanged.
 - **The full mix is a stem** (#933) — core no longer depends on `original_audio:`, a
   top-level manifest key this repo invented (#583) that the feedpak spec never had.
   The format already carried the pre-separation mixdown as a stem; feedpak 1.15.0
--- a/plugins/folder_library/screen.js
+++ b/plugins/folder_library/screen.js
@@ -895,6 +895,7 @@ function createFolderSurface(cfg) {
     var VIRTUAL_MIN = 200;   // below this, render everything — no behaviour change
     var VIRTUAL_BUFFER = 6;  // rows kept rendered above/below the viewport
     var _virtualCleanups = [];
+    var _virtualLists = [];   // repaint fns, one per live windowed list
 
     // Which slice of the list is on screen. Pure arithmetic — kept separate from
     // the DOM so it can be tested directly (see tests/virtual_list.test.js).
@@ -926,6 +927,7 @@ function createFolderSurface(cfg) {
     function _clearVirtualLists() {
         _virtualCleanups.forEach(function (fn) { try { fn(); } catch (_) {} });
         _virtualCleanups = [];
+        _virtualLists = [];
     }
 
     // Fill `list` with `songs`, windowed when the list is big enough to matter.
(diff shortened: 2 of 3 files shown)


byrongamatos:
@coderabbitai Both Major findings were real and are fixed in `3d02295`. Good catches — they were mine.

**`screen.js:985` — grid windowing on resize.** Correct. `perRow` and `rows` were captured once when the list was filled, but `paint()` also runs on resize, and resizing changes the column count. The window maths then sliced against the old count: wrong songs on screen, and padding sized for a row count the layout no longer had, so the scrollbar lied. `metrics()` now recomputes `perRow` / `itemH` / `rows` **together on every paint**, so the geometry cannot disagree with itself.

**`screen.js:987` — repaint on visibility change.** Also correct, and this one would have been very visible. A windowed list's contents are a function of its **position**; expanding or collapsing any section moves every list below it, so those lists kept the window computed at their old position and would show blank padding where songs should be until the user happened to scroll. Re-opening an already-populated section had the same flaw. Both toggles now call `_repaintVirtualLists()`.

On the second half of that finding — collapsed lists still paying `getBoundingClientRect()` on every scroller tick — `paint()` now bails early when the list is `display:none` or detached, and **forgets its last window** so re-showing repaints from scratch rather than short-circuiting on a stale memo.

**`screen.js:897` — CHANGELOG.** Added under `[Unreleased] / Changed`.

Tests: grid re-window on a column-count change, the padding+rendered=rows invariant at two different `perRow` values, and one that **pins the failure mode** — a mismatched `perRow`/`rows` pair must not silently look correct. 12/12. Re-validated the DOM glue in real Chromium with 50k rows (25–31 rows rendered, scroll height exact and constant). eslint clean; JS 1189/1189; pytest 2597 passed.

PR #972, opened 2026-07-14
feat(song-info): publish the playable stem list so stems can preload (fixes the 698ms freeze)

byrongamatos:
The stems plugin could only learn its stem list from the highway's WS `ready` — which arrives once the highway is **already up**. So it fetched, decoded, and then handed every stem's PCM to its audio worklet, copying the **whole song**, with the player on screen.

For a 4-minute 6-stem pack that is over **half a gigabyte of memcpy, in one frame, on the main thread**. Measured on a real load:

```
frame times (ms): ... 10.3, 9.3, 10.2, 10.3, 9.7, 697.9, 6.7 ...
                                              ^^^^^ right as the credits card appears
hottest: buildGraphFromBuffers [stems/src/main.js]
```

That is the *"the video pauses when the author appears"* report — the venue video's texture simply isn't drawn for 0.7s.

`GET /api/song/{f}?stems=1` now returns the same list (`[{id, url, default}]` + `full_mix_url`), so the plugin can start the whole load at `song:loading`, **before the highway and the venue are drawn**, where a stalled frame costs nothing. Nothing about the work changes; only *when*.

**Opt-in** via the query param, so the library's metadata calls — the hot path — pay nothing. Deliberately **not** cached: the metadata cache is a fixed-column table, and widening it would mean a schema migration plus a stale row for every already-scanned song, to cache what is a plain manifest read on an already-unpacked pack.

## The safety property

REST and the WS must publish the **same** list. If they disagreed, the plugin would preload a graph and then throw it away and rebuild — *strictly worse than not preloading*. So both now resolve `default` through one shared helper (`stem_default_on`, extracted from `load_song`), and the test **rebuilds the WS's payload from `load_song` and requires the REST helper to produce the identical list** — pinned against each other, not against a snapshot.

Also pinned: the mixdown is lifted **out** of the stem list (spec §5.3 — listing `full` beside the instruments would play the whole song *on top of* the stems) while staying reachable as `full_mix_url`; a single-`full` pack keeps it as its only playable stem; an unreadable pack yields an empty list rather than failing the request.

Full suite 2608 passed. Consumed by [feedBack-plugin-stems](https://github.com/got-feedBack/feedBack-plugin-stems/pulls).

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in `stems` parameter to `GET /api/song/{filename}` to return playable stem details (`id`, `url`, `default`) along with `full_mix_url` when available, matching the existing WebSocket “ready” payload behavior.
* **Bug Fixes**
  * Improved normalization of stem `default` values (including common off variants) for consistent playable-stem metadata.
  * Corrupt or unreadable song packs now return an empty stems list with `full_mix_url: None` instead of failing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   carry their gig log; instruments their gig count.
 
 ### Changed
+- **`GET /api/song/{f}?stems=1`** (new, opt-in) — returns the pack's playable stem
+  list (`[{id, url, default}]` + `full_mix_url`), the same list the highway's WS
+  `ready` sends. The stems plugin could only learn it from that WS message, which
+  arrives once the highway is already on screen — so it decoded and then copied the
+  whole song's PCM to its audio worklet with the player visible: over half a gigabyte
+  of memcpy in one frame for a 6-stem pack, a measured 698 ms freeze right as the
+  song-credits card appeared. With the list available at `song:loading` the plugin
+  does all of it before the highway is drawn. Built by calling `load_song` itself, so
+  it cannot drift from what the WS sends. Opt-in, so the library's metadata calls pay
+  nothing.
 - **Folder library renders only the songs on screen** (#965) — a song list used to
   render *every* song it held. On a flat 50,944-song library that was one `<div>`
   with 50,938 children and ~1.3 **million** DOM nodes (~4.2 GB of renderer memory),
--- a/lib/routers/song.py
+++ b/lib/routers/song.py
@@ -829,45 +829,49 @@ def post_song_gap_fill(filename: str, data: dict):
     return {"ok": True, "written": additions, "skipped": skipped}
 
 
-def _playable_stems_payload(song_path, filename: str) -> dict:
+def _playable_stems_payload(filename: str, dlc) -> dict:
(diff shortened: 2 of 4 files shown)


byrongamatos:
@coderabbitai the legacy full-mix finding is addressed in `98a2702`, and it was a real bug that would have hit most real libraries.

You were right: `load_song()` falls back to the deprecated `original_audio:` key when a pack has no reserved `full` stem — which is **every pack written before feedpak 1.15.0**. My payload reimplemented the full-mix rule from `extract_meta` and returned `None` for those, so REST would say "no full mix" while the WS said there was one.

Worse than a wrong field: the stem *signature* would still match, so the plugin would preload a graph **without** the pristine mix and never rebuild — unity playback silently downgrading to the lossy stem recombination. That is exactly the drift this PR claims to prevent, and my test had a hole (I only covered packs carrying a `full` stem).

The fix is structural rather than another special case: `_playable_stems_payload` now **calls `load_song` itself** — the same function the WS uses — whose `LoadedSloppak` already carries the partitioned stems and the resolved full mix, and builds the URLs exactly as `ws_highway` does. Drift is now impossible by construction instead of by agreement. `extract_meta` is reverted to its original shape (it never needed to change).

Tests rewritten to compare against `load_song` for a reserved-`full` pack, a **legacy `original_audio` pack** (the case that was broken), and a single-`full` pack.

PR #979, opened 2026-07-15
perf(scan): skip full library re-stat when the tree is unchanged

byrongamatos:
## Problem

Every startup scan globbed the whole DLC tree twice (`*.feedpak`, `*.wem`) and `stat()`'d every file to detect changes — ~100k filesystem round trips on a 50k-song library. On a slow NTFS-3G FUSE mount this is the "big drive churns on every launch" report.

## Fix

Adds/removes/renames of songs all bump the mtime of the **containing directory** (verified on the target mount). So after a full pass we persist `{reldir: mtime_ns}` for every library dir to `scan_dir_signature.json` (keyed by DLC path so switching libraries never matches a stale signature). The next scan re-stats **only those dirs** — a handful vs 100k ops — and skips the entire listing/stat pass when none changed.

- Signature recorded at both full-pass completion points (nothing-new and scan-complete).
- Seeding runs before the fast-path check and is idempotent, so it doesn't perturb mtimes on a settled library.

## Blind spot + escape hatch

A pack rewritten **in place under the same name** bumps the file mtime but not its dir's — the fast path can't see it. Rare for a song library (you add/remove packs, you don't rewrite them under the same name). The manual Refresh covers it: `/api/rescan` and `/api/rescan/full` now pass `force=True`, which skips the fast path and always does the full pass. `force` threads through `kick_scan` → `_scan_runner` and coalesces like the existing rescan-pending flag.

## Tests

`test_dir_signature_fast_path_skips_unchanged_tree`: full pass records signature → unchanged tree takes fast path (zero extraction) → new file bumps dir mtime → full pass picks it up → `force=True` bypasses the fast path. Scan + career + settings suites green (169 passing).

## Summary by CodeRabbit

- **New Features**
  - Added a directory-content signature to speed up library startup scans when nothing has changed.
  - New or modified library files still trigger full scanning when required.
  - Manual and full rescan actions now force a fresh scan, bypassing the fast path.
- **Bug Fixes**
  - Improved handling of updates made “in place” by ensuring rescans refresh correctly even when directory timestamps don’t change.
- **Tests**
  - Added coverage for the directory-signature fast path, including forced bypass and directory-form song tracking.

### DIFF
--- a/lib/scan.py
+++ b/lib/scan.py
@@ -51,6 +51,87 @@ from scan_worker import _relpath, _scan_one
 
 log = logging.getLogger("feedBack.scan")
 
+import json
+
+
+# ── Directory-signature fast path ─────────────────────────────────────────────
+#
+# A startup scan globs the whole library twice (*.feedpak, *.wem) and stats every
+# file to detect what changed. On a 50k-song library that lives on a slow mount
+# (an NTFS-3G FUSE volume here) it is ~100k filesystem round trips every launch —
+# the "big drive churns on every startup" report.
+#
+# But adds / removes / renames of songs all bump the mtime of the DIRECTORY that
+# holds them (verified on the target NTFS-3G mount), and so does the addition of
+# a subdirectory (a new entry in its parent). So after a scan we record every
+# library directory and its mtime; on the next scan we re-stat ONLY those
+# directories (a handful, vs 100k file ops). If none changed, the file set is
+# unchanged and the whole listing/stat pass is skipped.
+#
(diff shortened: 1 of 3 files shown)


byrongamatos:
@coderabbitai review

byrongamatos:
Fixed in 20a5074 — `_library_dirs` now records a directory-form song's own dir (via `f.is_dir()`), so an in-place change inside a loose-song folder / directory bundle invalidates the fast path. File-form sloppaks are unaffected. Added a unit test. @coderabbitai review

PR #984, opened 2026-07-16
fix(gp2rs): write arrangement XML as UTF-8 so GP import survives non-ASCII metadata

ChrisBeWithYou:
## The bug

Importing a Guitar Pro file whose metadata contains a non-ASCII character fails on **Windows** with:

```
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line N, column 22
```

and the whole import returns HTTP 500. Repro'd with a real file whose album name is `Chrysalis©1982` — the generated `<albumName>` line is exactly where it dies.

## Root cause

The GP→arrangement-XML writers persist their output with `Path.write_text(xml_str)` and **no explicit encoding**. `Path.write_text` uses the platform default text encoding — **cp1252 on Windows** — so `©` (U+00A9) is written as the lone byte `0xA9`. `parse_arrangement` then reads the file back as UTF-8 (expat's default), where `0xA9` is an invalid start byte → parse error.

CI runs on Linux, where the default is UTF-8, so the bug is invisible there — which is why it shipped.

## Fix

Pin `encoding="utf-8"` on all three arrangement-XML writes: `lib/gp2rs.py` (1) and `lib/gp2rs_gpx.py` (2).

## Test

`tests/test_gp2rs_xml_encoding.py` — because a plain functional test passes on Linux CI even with the old code (UTF-8 default), the regression is pinned at the source-contract level (no bare `write_text(xml_str)`; the UTF-8 form is present), plus a round-trip that a `©` album name written as UTF-8 parses back intact. Converter suites: 215 passed.

## Context

Found while dogfooding the Arrangement Editor's GP import on Windows. Independent of the editor plugin — this is core `lib/`.

https://claude.ai/code/session_01EBQCHCNA81E9tHmSDHSe2Q

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Guitar Pro imports on Windows failing with non-ASCII song metadata.
  * Arrangement XML generation now reliably writes output using UTF-8 to prevent malformed XML import errors.
* **Tests**
  * Added regression tests to ensure generated arrangement XML is always written with UTF-8 and that non-ASCII album metadata round-trips correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -229,6 +229,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   engine (`app.js`, `highway.js`, `playSong`, `showScreen`, the capability registry).
 
 ### Fixed
+- **Guitar Pro import no longer fails on non-ASCII song metadata (Windows).**
+  The GP→arrangement-XML writers wrote their output with `Path.write_text()`
+  and no explicit encoding, so on Windows (cp1252 default) a metadata
+  character like the © in an album name ("Chrysalis©1982") was written as a
+  lone `0xA9` byte — invalid UTF-8 — and import died with
+  `not well-formed (invalid token): line N, column 22`. All three arrangement
+  XML writes now pin `encoding="utf-8"`.
 - **3D Highway: the lane stops at the hit line** (#991) — the highway lane, its
   dividers, and the fret boundary extension lines ran `BEHIND` seconds *past* the
   hit line toward the player. Nothing is ever drawn in that strip (notes and chord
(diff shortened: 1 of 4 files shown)


ChrisBeWithYou:
Review pass complete. Found and fixed one regression-test gap in 1e8302d: the original regex could miss incorrectly encoded or differently formatted write_text(xml_str, ...) calls, and the round-trip test leaked its temporary directory. The test now inspects every arrangement XML write via the Python AST, pins all three writer paths to UTF-8, and uses pytest's managed temp directory.

Verification: 234 converter/notation tests pass on Windows, git diff --check is clean, and all PR checks (including CodeRabbit) are green. A second review pass found no further issues.

Issue #56, opened 2026-06-20
Background rescan migration for songs scanned before the `stem_ids` / `tuning_name` / `...


byrongamatos:
> _Restored from `slopsmith/slopsmith#221` — original issue, opened by @byrongamatos on 2026-05-09._
> _[restored-from: slopsmith/slopsmith#221]_

_From `specs/001-slopsmith-platform/tasks.md` — generated retrospectively by spec-kit._

Background rescan migration for songs scanned before the `stem_ids` / `tuning_name` / `tuning_sort_key` columns existed — `CHANGELOG.md` "Migration notes" mentions Full Rescan recovery, but a one-shot lazy backfill would be friendlier

---
See `specs/001-slopsmith-platform/spec.md` and `specs/001-slopsmith-platform/analyze.md` for context.
This issue was filed in bulk; close as `wontfix` if no longer relevant.

PR #997, opened 2026-07-17
feat: instruments as plugins — extract hardcoded instrument knowledge into plugin-based registry

j0b333:
## What

Extracts all hardcoded instrument knowledge (guitar, bass, drums, keys) from ~15 scattered Python and JS files into a plugin-based registry. Guitar, Bass, Drums, and Keys now ship as bundled instrument plugins (`plugins/instrument_<name>/plugin.json`) each defining their own tunings, roles, arrangement mappings, string/key counts, detection strategy, and icon. Community instruments drop in as new `type: "instrument"` plugins with zero code changes.

### Backend
- `lib/instruments.py` — registry class with schema validation, offset-to-MIDI computation
- `GET /api/instruments` — serves all registered instrument definitions
- Plugin loader recognizes `type: "instrument"` and auto-registers definitions
- `lib/tunings.py` — profiles, tuning validation, preset catalogs from registry
- Arrangement routing (`ws_highway.py`) uses role flags/names from registry
- Progression `instrument_for_arrangement` accepts registry, returns None for unknown
- Settings API validates against registered instrument IDs
- `/api/tunings` populates both tuningMidis and tunings (fixes tuner sync)
- Non-stringed instruments skip string_count/tuning validation

### Frontend
- Dynamic instrument selector with per-instrument icons, roles, key counts
- Per-role mastery % on library grid (not song-wide max)
- Hover overlay on cards showing all arrangement scores
- Per-instrument preferred highway in Settings → Instruments
- Auto-filter library by current instrument arrangement roles
- 'Auto-filter by instrument' toggle in Settings → Gameplay
- Instruments settings tab with editable tunings, names, string counts
- Working-tuning capability reads from registry
- Score attribution fix: `song:ready` resolves arrangement index

## feedpak surface

- [ ] This change touches a manifest key, file, or directory that must land in feedpak-spec first
- [x] This change does NOT touch feedpak-spec territory

## Checklist

- [x] Tests pass — 132 relevant tests green
- [x] JS syntax validates — `node --check` passes on all modified JS
- [x] No `print()` calls in modified Python
- [x] Logging follows `logging.getLogger("feedBack.*")` convention
- [x] Plugin IDs use snake_case, JS uses camelCase, Python uses flat imports
- [x] DCO sign-off: I certify this contribution is my own work and I have the right to submit it

Full changelog in `CHANGELOG.md`.

## Summary by CodeRabbit

* **New Features**
  * Added Instruments-as-plugins with dynamic instrument selection (stringed and non-stringed) and role-aware arrangement routing.
  * Added **GET /api/instruments**, a new **Instruments** settings tab (per-instrument roles, patterns, tuning/string/key overrides, and preferred highway), plus automatic song library filtering by instrument.
  * Added per-arrangement best-accuracy reporting via **/api/stats/best-by-arrangement**.

* **Bug Fixes**
  * Improved tuner/instrument tuning sync and adjusted arrangement accuracy attribution.
  * Fixed arrangement-filtering behavior (including an SQL edge case) and removed the default arrangement dropdown in favor of **Auto-filter by instrument**.


j0b333:
@coderabbitai review

j0b333:
@coderabbitai review

j0b333:
@coderabbitai review

j0b333:
@coderabbitai review

j0b333:
I really made sure to go over all my code and add all the docstrings, I know they are just warning but cant help but feel this low coverage was already there. Any feedback? Also the acoustid_api_key is also not something I added. This should probably be resolved but not sure if this PR is the place.

j0b333:
Some images to show the changes made.

image
image
image
image
image

byrongamatos:
Hey thank you for that, it shows you put a lot of effort into it. Let me go through this in detail, see if there are any unintended side-effects

jphinspace:
Ideas/suggestions
1. Allow multiple tunings/names/stringcounts/etc per instrument, eg two players might want to play together splitscreen, one lead one alt lead, one with 6 string one with 7
2. This might be a good place for the string color selector/drum pad color selectors to live so they can be reused across different note highway visualizations (highways can still provide the default if not configured here)
3. Also could include storing the highest fret number on each instrument so highways won't show/score unplayable notes

j0b333:
@jphinspace 
> Ideas/suggestions
> 
> 1. Allow multiple tunings/names/stringcounts/etc per instrument, eg two players might want to play together splitscreen, one lead one alt lead, one with 6 string one with 7
> 2. This might be a good place for the string color selector/drum pad color selectors to live so they can be reused across different note highway visualizations (highways can still provide the default if not configured here)
> 3. Also could include storing the highest fret number on each instrument so highways won't show/score unplayable notes

Hey, good suggestions and sort of the direction I was already thinking moving next. Once this gets merged I was thinking on working fleshing out the multiplayer a bit.

1. What I was imaging is moving towards more than just the default profile so people can load their own profile (or guest) and join a game, so just we are used to in other gh/rocksmith type games etc. Then have every player select their input source and instrument. And do multiplayer that way. I think it would be much more intuitive than setting up multiple instrument settings in the single instrument profile.
2. I was thinking of what should live in the specific instrument settings and what should live in their own plugin settings. Personally I feel that things like string color should be part of the highway settings. Maybe per profile. I think it deserves a little more thought, if you have any suggestions keep them coming though. 
For the drums I would agree it would make more sense moving the drumkit configuration from the highway into the instrument settings and have the highway consume those settings from the instrument settings instead. But I did not want to mess around with how the drums were setup without fully understanding and maybe see if the original creator has good reason not too etc.
3. I agree it would make sense to have the fret count next to string count, is the functionality for dealing with notes above the highest fret already integrated in the highways though? So yeah I agree it should go here but also tried to keep away from changing the highways in this change as well seeing as I did not dive into their respective codes yet. Tried to just make changes that integrate with the current setup.

j0b333:
@byrongamatos 
I see quite a lot of changes have been merged since this PR with a few having quite a bit of overlap. I am currently working on resolving and integrating my solution within the current version of main. Let's hope there won't be a gazillion changes before the next PR haha but the more people working on this project the better.

j0b333:
@byrongamatos 
Reimplemented my changes but from the current state of the main branch. While at it I did add fret count settings for stringed instruments (not used anywhere yet but now configurable) and a skeleton for a vocals plugin because I believe that finishes what is promised for instruments in the base FeedBack.

The content decisions when merging into upstream main were:

songs.js — built on upstream's version
Upstream had already built perspective-aware tuning (instrument= query param, bass/rhythm tuning columns, playable-without-retuning filter). I ported my registry-based features (~200 lines) onto that base rather than reimplementing 2000 lines of theirs.

Registry as source of truth for instruments
Upstream's libInstrument() only knows guitar-lead, guitar-rhythm, and bass. My instruments-as-plugins system supports any instrument — drums, piano, vocals, community plugins. I kept my window.feedBack._instruments registry as the authority and routed upstream's perspective-aware patterns through it.

Per-role accuracy with hover overlay — kept mine
Upstream shows a single song-wide percentage. Selecting "Rhythm" would display Lead scores. Mine resolves the correct arrangement via _matchingArrangementIndex() and fetches /api/stats/best-by-arrangement. The hover overlay showing all arrangements' individual accuracies was also unique to my implementation, so I kept it.

Auto-filter toggle — kept mine
Upstream derives filtering from the active profile with no off switch. Mine has an explicit auto_filter_instrument setting toggle. I adopted upstream's instrument= query param for server-side tuning correctness, but kept the user-facing toggle.

Dynamic arrangement filter pills — kept mine
Upstream hardcoded ['Lead','Rhythm','Bass','Combo','Vocals']. Mine derives labels from the instrument registry — adding a new instrument plugin automatically populates the filter drawer.

Tuning chip position — kept mine
Flush top-left (top-0 left-0 rounded-br-md) with hover fade-out. Cleaner than upstream's 2px offset; fading prevents the chip from obscuring action buttons.

Tuning offset detection — combined both
My original approach tried song.tuning_offsets first. But older songs with empty tuning_offsets in the DB needed upstream's name-based fallback chain. The final chain: shown.tuning_offsets → perspective-aware offsets → raw name → legacy tuning field. Four sources, no missed matches.

Bass chip matching — fixed upstream's bug
Two interrelated bugs in upstream: shownTuningOffsets gated on the name field rather than the offsets field (skipping populated bass_tuning_offsets when bass_tuning_name was empty), and chipIsBass required a native bass chart name (failing for inferred cards). Both caused bass chips to show permanent amber. Fixed by gating on the offsets field directly and marking any bass-player chip as bass.

Double reload on instrument switch — fixed
Both instrument:changed and working-tuning-changed fired reload() on instrument switch, racing state mutations. Fixed by setting _lastRenderInstrument early so the tuning-changed path sees no change and only calls decorateTuningChips.

Chip logic made registry-aware — replaced hardcoded bass/lead
Was: chipIsBass hardcoded, truncation always to 4, coverage always 'Bass' | 'Lead'. Now: _activeInstrument() reads from the registry, truncation uses the instrument's string_count, chip carries data-tuning-instrument for any future instrument. data-tuning-bass kept for backward compat with the existing tuner plugin.

Bass pick/fingered arrangement names — added
GP imports commonly name bass parts "Pick Bass" or "Fingered Bass". Upstream's exact-match name path only had ["Bass"]. Added ["Pick Bass", "Fingered Bass", "Finger Bass", "Picked Bass"]. The flag path (path_bass=True) already caught RS2014 imports.

Missing function, encoding corruption — fixed
renderDrawerIfOpen was lost during the port (crashed on instrument switch). And instruments-settings.js had mojibake (ΓåÆ instead of →, ΓöÇ instead of ─) from a wrong-encoding save. Both restored.

j0b333:
@coderabbitai review

gionnibgud (commented):
Reviewing from the keys/instrument-foundations side, since this lands right on
that surface. Short version: the direction is right and most of this I'd want to
keep. One design question I think needs a maintainer call before it merges, one
coordination flag, two concrete registry bugs, and one packaging ask.

Thanks for taking this on — the three-disagreeing-guess-tables problem is real
and has bitten several of us.

## The design question: name-based routing as a plugin contract

`roles[].arrangement_names` makes the arrangement's **display name** the routing
key, and puts it in `plugin.json` — a public, community-facing contract.

That collides with **[feedpak-spec#56](https://github.com/got-feedBack/feedpak-spec/issues/56)**,
which is open (ON HOLD, prototyping editor-side first) and exists specifically to
decouple instrument routing from arrangement name. Its problem statement is the
same three tables this PR consolidates — so we agree on the diagnosis; the
question is the cure. #56 proposes a per-arrangement `roles` list in the pack;
this PR keeps the name as the key and moves the guessing into manifests.

Consolidating three tables into one is a clear improvement either way. My concern
is specifically the **manifest surface**: once community instrument plugins ship
`arrangement_names`, moving to a spec-driven key later breaks them. That makes
this the expensive half to get wrong.

Concretely, the normative and the cosmetic currently share one namespace
(`lib/progression.py`):

```python
if name in role.get("arrangement_names", []):
    return inst["id"]
if arr_type in role.get("arrangement_names", []):
    return inst["id"]
```

`type` is a spec field (§5.2, and `type: "drums"` is normative as of 1.17.0);
`"Drum Kit"` and `"Synth Lead"` are display strings. Matching both against one
list works today only because the type tokens happen to appear in the name lists.
If a future spec minor adds a normative `type` value, routing silently misses it
until every plugin manifest adds it as a "name".

Two options that would both address this without shrinking the PR much:

1. Separate the keys: `arrangement_types` (matched against spec `type`, exact)
   from `arrangement_names` (matched against display name, fuzzy/fallback). Types
   win. This is additive and leaves room for #56 to land without a breaking
   manifest change.
2. Or land the registry as-is but mark `arrangement_names` **provisional** in
   `docs/instrument-plugins.md` — explicitly "subject to change pending #56, do
   not depend on it in third-party plugins yet."

I'd prefer (1), but either resolves my objection. Flagging for a maintainer call
since it's a spec-adjacent contract, not purely a core decision.

## Coordination flag: `instrument_vocals`

Vocals-as-a-typed-part is currently fenced to a coordinated instrument-pathways
FEP by an explicit agreement between the keys-foundations and song-editor lanes,
and there are vocal sidecar PRs in flight separately (#1031, #1037, #1038). This
PR adds a third, independent vocal typing (`detect_strategy: "pitch"`, vocal role
names).

Nothing wrong with the definition itself — it's the same shape as the others. But
I'd suggest **dropping `instrument_vocals` from this PR** and letting vocals
enter through the coordinated FEP, so we don't end up with three definitions of
what a vocal part is. The registry is the right home for it *eventually*; it's
the timing that's awkward.

## Two registry bugs

**1. No cross-instrument name-collision detection.** Duplicate role ids *within*
one instrument raise `ValueError`, but two instruments claiming the same
arrangement name coexist silently, and the winner is decided by dict insertion
order:

```python
def instrument_id_for_arrangement(self, arr_name, arr_flags=None):
    for inst in self._instruments.values():   # insertion order
```

Registration order is plugin load order, which `CLAUDE.md` documents as
alphabetical by directory. So a plugin at `plugins/instrument_aardvark/` that
claims `"lead"` silently outranks `instrument_guitar` for every Lead arrangement
in the library, with no warning anywhere. Suggest rejecting (or warning + first-
wins on) a name already claimed by another instrument.

**2. Silent overwrite of an existing instrument id.**

```python
self._instruments[inst_id] = normalized
```

A community plugin declaring `instrument.id = "keys"` replaces the bundled keys
definition outright — no warning, no first-wins. Given instrument definitions
now drive routing, tuning validation and detection strategy, that's a footgun
worth closing. First-wins plus a `log.warning` would match how the loader treats
other plugin-supplied registrations.

## Packaging ask: split the reindentation

`static/v3/songs.js` is +4682/−4455 in a **single hunk**. Diffing it with
whitespace normalized, 3448 of the added lines are identical after stripping —
roughly three-quarters of the diff is reindentation, with only ~180 genuinely
new lines inside it. The real change isn't reviewable in that form, and it'll
make every future `git blame` on that file dead-end here.

Could you split it into a whitespace-only commit and a functional one? Same
final tree, and it makes the substantive part readable. Same goes, to a lesser
extent, for the features bundled in beyond instrument extraction (per-role
mastery %, hover overlay, auto-filter toggle, score attribution, the new
`/api/stats/best-by-arrangement`) — those are each defensible, but they widen
the blast radius of a PR that's already large.

## Not blocking, minor

- `plugins/instrument_piano/` declares `instrument.id: "keys"` with label
  "Keys" — the id is right (core uses `keys` throughout: `keys_highway_3d`,
  `data/progression/paths/keys.json`), so consider renaming the directory to
  `instrument_keys` for consistency.
- Heads-up that this touches `lib/routers/ws_highway.py` in the arrangement-
  routing region, where there's rig/tone-binding work in flight — happy to
  handle the merge on my side, just flagging it.

j0b333:
@gionnibgud 
Thanks for this extensive review of the change. I think you brought up quite a few valid concerns and bugs. I think I agree with almost of your FeedBack and will try to fix this for so we can get this pushed. 
About the Vocals I see what you mean, it was meant as a skeleton for other vocals branches to merge into with the new plugin architecture if they wanted to but at the same time I am just adding a plugin with 0 functionality as off now and cluttering the process.
Also for the name based routing I will try if moving it to type will break anything. Let me see what I can do getting ready for the spec change while also allowing it to work right now.

Let me also see what I can do splitting some of the features in their won PR. Used to working with small teams where taking on bigger changes was fine. But that does not work with 10s of PRs per week haha. I will try and make my changes more manageable in the future.

j0b333:
Hey @gionnibgud

Design question — arrangement_names vs spec types. I went with your option 1. Added arrangement_types alongside arrangement_names as a separate field. Types match exactly against the spec type value, names are the fuzzy display-name fallback. Types win on collision. Updated all four bundled manifests, the registry matching logic, progression, and docs. It's fully additive — existing plugins without arrangement_types keep routing through names. This leaves room for feedpak-spec#56 without a breaking manifest change. I also display the types read-only in the settings panel's role cards (e.g., "Lead (lead) (default)").

Coordination — dropped instrument_vocals. Removed the plugin directory entirely and reverted the vocal routing line from the hardcoded fallback in progression.py. Vocals still work via the pre-existing fallback. Will let the FEP land it properly when ready.

Minor naming — instrument_piano → instrument_keys. Renamed the directory to match the "keys" id used everywhere in core.

Registry bug 1 — cross-instrument collision. InstrumentRegistry.register() now iterates all already-registered instruments when a new one comes in and warns if any arrangement names or types overlap, identifying both the new and conflicting plugin IDs. First-wins remains the policy.

Registry bug 2 — duplicate id. Changed from raising ValueError to first-wins + log.warning with both plugin IDs, so community plugins can't silently displace bundled instruments.

Packaging — songs.js split. Split into two commits on the fix branch (PR #2 targeting this feature branch):

- Commit 5a6772a (chore: normalize indentation in songs.js) — takes main's songs.js and reindents it to match the feature branch's formatting. Zero functional changes. Verified with git diff -w against main: no output, meaning after ignoring whitespace the files are identical. This isolates the formatting noise so the real changes are reviewable.
- Commit dd1e183 (feat: instrument-aware library features) — applies all the instrument-aware functional changes on top of the reindented file: per-role mastery %, hover overlay, tuning perspective, auto-filter toggle, grid cursor navigation, and the instrument:changed / working-tuning-changed event handlers. git diff -w between the two commits shows only the genuinely new logic.

Both commits cancel to a net-zero diff against the feature branch — same final content, just layered so reviewers can step through formatting vs. logic separately.

j0b333:
@gionnibgud  

anyone still planning to look into this?

Issue #118, opened 2026-06-20
Installer flagged by Windows Defender / Smart App Control


byrongamatos:
> _Restored from `slopsmith/slopsmith#519` — original issue, opened by @jphinspace on 2026-05-15._
> _[restored-from: slopsmith/slopsmith#519]_

Version: 0.2.7
OS: Windows 11
Workaround: 1. Disable Smart App Control. 2. When the Windows Defender popup appears, click "More Info" and then the "Run Anyway" button will appear

It looks like Slopsmith.Setup.0.2.7.exe doesn't show show anything in Properties -> Digital Signatures, probably just need to sign the exe (or include the workaround in the README)

byrongamatos:
_@byrongamatos commented on 2026-05-15:_

We are waiting on approval for a signing certificate. This warning will be gone when we get it.

byrongamatos:
_@byrongamatos commented on 2026-05-28:_

Update: we're in contact with [SignPath.io](https://signpath.io) to obtain a code-signing certificate for the Windows builds (they provide free certificates + signing infrastructure for open-source projects). Once that's set up, the installer and executables will be Authenticode-signed and the Windows Defender / Smart App Control warning will go away.

In the meantime the workaround in the issue stands: on the SmartScreen/Defender popup, click **More info → Run anyway** (and, if Smart App Control is on, the unsigned binary may need it temporarily disabled).

PR #998, opened 2026-07-17
Enable Linux nightly AppImage auto-update in the System settings UI

mhglover:
## Summary
- Fixes the Settings → System "App updates" panel so it actually works on Linux: the channel dropdown no longer gets permanently disabled the moment the desktop bridge reports `unsupported` (which is the normal state whenever the channel isn't Nightly on Linux) — it stays enabled so the user can switch to Nightly, which is the only way out of that state.
- Adds Nightly to `APP_UPDATE_CHANNELS` and the channel `<select>` — previously missing entirely, so Linux self-update couldn't be reached from this UI at all.
- Shows live download progress ("Downloading update… N%") and an explicit state machine on the Check/Restart button, instead of a frozen "Checking…" during the ~1.5GB background download.
- Two real UI bugs found via on-device Export Diagnostics captures, both fixed: (1) `setupAppUpdates()` was re-syncing the channel on every Settings re-render, which could stomp an in-flight download's state; (2) the very first channel sync on page load raced against the initial status render and could get stuck showing "unsupported" even after the channel was correctly set.
- Routes update-flow logging into the existing `diagnostics.js` console-capture + `contribute()` snapshot API, so a user's existing "Export Diagnostics" button now captures the full update decision trace for future debugging — no new UI or log file.

Companion PR in feedBack-desktop (the underlying update engine): got-feedBack/feedBack-desktop#118

## Test plan
- [x] Every renderer change verified with a standalone browser harness (mocking `window.feedBackDesktop.update` and `window.feedBack.diagnostics`) reproducing the exact bug sequence before confirming the fix, for each of the fixes above.
- [x] Verified end-to-end on a Steam Deck: channel switch → check → live download progress → restart button → relaunch onto the new build, confirmed via a real Export Diagnostics capture showing a clean, fully-accounted-for trace.

### DIFF
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -209,7 +209,7 @@ export function setupWindowOptions() {
     }
 }
 
-export const APP_UPDATE_CHANNELS = ['stable', 'rc', 'beta', 'alpha'];
+export const APP_UPDATE_CHANNELS = ['stable', 'rc', 'beta', 'alpha', 'nightly'];
 
 export let _appUpdatesWired = false;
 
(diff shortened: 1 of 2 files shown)


mhglover:
Closing to clean up the commit history before requesting review — will reopen once it's squashed into a clean commit.

PR #1000, opened 2026-07-18
Add new chart-transform plugin capability

### LINKED ISSUE #952: Feature request: let plugins transform chart data (notes/chords) before rendering and scoring
Plugins can replace the highway's rendering (`setRenderer`), decorate it (overlays, draw hooks), and judge it (note-state provider), but cannot change the displayed chart data itself. There is no supported way to remap, add, or remove notes between the WebSocket load and the consumers of that data.

Today, a plugin whose whole purpose is note mapping is forced to fork an entire renderer to substitute its remapped notes at draw time (eg my plugin https://github.com/jphinspace/feedBack-plugin-chart-retuner ). This has two structural problems:

- The mapping exists only inside that one forked viz. The built-in 2D highway, 3D highway, jumping tab, tab view, and splitscreen panels never see it. The fork has to be manually re-synced against upstream on an ongoing basis.
- Scoring doesn't follow. note_detect reads the highway getters, which serve the original chart, so judgments are based on positions the player is deliberately not playing.

The request is for a supported hook for a plugin to transform the loaded chart, with the result flowing to every downstream consumer (visualizations, overlays, splitscreen panels, scoring, etc). A transform plugin should ship chart math and a settings screen, without having to worry about the highway rendering.

jphinspace:
## What

Apologies if this change is pre-emptive, since I know you all have a lot on your plate at the moment!

The idea behind this change is to expose more chart data to plugins, including access & modification of the placements of notes, strings, hand positions, and other chart data, all without modifying the original.

The motivation for this change was to unblock my alternate-tuning plugin (shift all the notes to a different tuning) but it could also be used for things like chord simplification (alternate voicing or root-notes-only), alternative dynamic difficulty implementations, new practice modes, and so on.

This would remove the need for these types of plugins to have to fork and maintain their own note highways/renderers. This change also provides data for note-detection to use to support the transformed notes.

Resolves https://github.com/got-feedBack/feedBack/issues/952.

## feedpak surface

- [X] This PR does **not** change how the app reads/writes feedpaks (manifest keys, pack files, folder layout)
- [ ] …or it does, and the spec change landed first via the [FEP process](https://github.com/got-feedback/feedpak-spec/blob/main/CONTRIBUTING.md) — FEP / spec PR: `got-feedback/feedpak-spec#___` (once it merges, re-run this PR's checks and the gate goes green)

## Checklist

- [X] `CHANGELOG.md` `[Unreleased]` updated (user-visible changes)
- [X] Tests added/updated for new behaviour
- [X] Commits are DCO signed off (`git commit -s`)

## Summary by CodeRabbit

* **New Features**
  * Added chart-transform support for plugins, allowing charts to be adjusted before rendering and scoring.
  * Transformations apply across standard, custom, and splitscreen views, with provider selection persisted between sessions.
  * Updated chart data, metadata, templates, and note timelines are reflected consistently in supported renderers and controls.

* **Bug Fixes**
  * Failed or invalid transformations safely retain the original chart.
  * Improved 3D highway handling of transformed tuning and capo information.

* **Documentation**
  * Added guidance for configuring, managing, and troubleshooting chart-transform providers.


jphinspace:
Outside diff range comments (1) intentionally skipped:

`The identity-based cheap key is the documented contract, not an oversight: bundle array fields are specified as swap-reference-on-change and read-only (CLAUDE.md's bundle rules; the chord render cache and the merge caches in highway_3d all key on identity the same way). No core path mutates a tuning array in place, and since the hardening pass deep-clones staged transform output, even a misbehaving provider can't reach the staged _xfTuning after the fact. Guarding against an in-contract-violating mutation with a per-frame content scan would contradict the zero-allocation purpose of the fast path and diverge from every sibling cache in the codebase.`

jphinspace:
@coderabbitai review

PR #1008, opened 2026-07-19
feat(highway_3d): background controls in the player chrome

Elit3d:
## What

Adds a background picker to the player's **Plugin Controls** popover, so the 3D Highway background can be changed mid-song without leaving for Settings.

**Controls:** style dropdown, Reactive toggle, Intensity slider. All write through the existing `h3dBgSet*` setters — no new persisted state, no new defaults. The style list is generated from `BG_STYLE_IDS`, and changes sync both ways with the Settings page.

**Greyed-out states** — controls the active style ignores are disabled with a reason on hover, driven by a per-style capability table (`_PC_USES`) derived from the `BG_STYLES` bodies:

| Style | Intensity | Reactive |
|---|---|---|
| Particles / Silhouettes / Lights / Geometric | ✅ | ✅ |
| Custom image | ✅ | — |
| Custom video · Butterchurn · Off | — | — |

**Notes**
- The control is refcounted rather than per-renderer-instance, since the settings it writes are global — multiple instances share one control instead of mounting one each. It unmounts when a non-3D renderer is selected.
- Intensity writes on release rather than per drag step, matching `settings.html` and avoiding a full scene rebuild per step.
- Styled inline using resolved `tailwind.config.js` tokens — no new Tailwind classes, no `assets/plugin.css` rebuild.
- `settings.html` is unchanged.

**Tests** — `plugins/highway_3d/tests/background_control.test.js` (13 cases) covers the refcounting, teardown, remount after a chrome rebuild, one-write-per-drag, and the per-style greying for all eight styles. Its expectations are derived from the `BG_STYLES` bodies rather than read from `_PC_USES`, so it can't pass by asserting the table equals itself.

The multi-instance path is covered by tests with stubbed instances only — it isn't verified against splitscreen, whose visualizer doesn't currently work (unrelated to this change; same behaviour on an unmodified build).

## feedpak surface

## Checklist

- [ ] Commits are DCO signed off (`git commit -s`)

## Summary by CodeRabbit

* **New Features**
  * Added 3D Highway player “Background” controls in the Plugin Controls popover, enabling mid-song background switching.
  * Includes style selection, Reactive toggle, and intensity slider; UI stays synchronized with the Settings page.
  * Unsupported background options are greyed out with hover explanations.
* **Bug Fixes**
  * Improved reliability of mounting/remounting the control during player/UI DOM changes, with safe acquisition/release across renderer lifecycle.
* **Tests**
  * Added automated coverage for sync behavior, refcounted multi-instance mounting, inert disabled controls (including hover reasons), and correct intensity update timing.


ChrisBeWithYou:
# Notes on feedBack#1008 — `feat(highway_3d): background controls in the player chrome`

Post-merge review notes for @Elit3d. Line numbers refer to `plugins/highway_3d/screen.js`
on `main` as of the merge (the `_pc*` block is ~4037–4360).

Context for where I'm coming from: I maintained the standalone Butterchurn fork of this
renderer before it was merged into core (got-feedBack/feedBack#597), so most of what's
below is about the seams that block touches — the `butterchurn` style, the venue override,
and per-panel background scoping.

First: this is a good change and the commenting is unusually careful. The
`_pcReactiveWrap` / `pointer-events:none` trick for showing a tooltip on a disabled
control, and the `change`-not-`input` choice on the intensity slider with the
`_bgRebuild()` reasoning spelled out, are both the right calls for the right reasons.
Everything below is follow-up, not a re-litigation of the design.

---

## 1. Venue override desyncs the control (real bug)

**This is the one I'd fix first.**

`_pcSync()` (4153) reads the *stored* style:

```js
const uses = _PC_USES[_bgReadSetting(null, 'style')] || { intensity: true, reactive: true };
```

But what's actually rendering is `_bgEffectiveStyleId()` (~8422):

```js
function _bgEffectiveStyleId() {
    return _venueSceneOverride ? 'venue' : bgStyleId;
}
```

`venue` is deliberately kept out of `BG_STYLE_IDS` (see the comment at 2402–2406) — it's an
internal effective style reached only through `h3dVenueSceneSetActive(true)` from the
viz-picker Venue flow. So while Venue is active:

- The dropdown shows the *stored* style (say "Particles (drifting)") while the user is
  looking at the venue scene.
- Intensity/Reactive are greyed per `_PC_USES.particles` — i.e. both enabled — but they're
  driving a style that isn't mounted.
- Picking a different style from the dropdown writes `style`, fires `_bgRebuild()`, and
  `_bgMountStyle` resolves to `venue` again. **The dropdown looks broken.**

Compounding it: `h3dVenueSceneSetActive` emits `_bgEmitChange('venueScene')` (2988–2996),
and `_pcListener` (4289) doesn't include `'venueScene'` in its key filter — so the control
doesn't even resync when Venue toggles on or off.

Two things to decide:

1. **Minimum**: add `'venueScene'` to the `_pcListener` key list so the control at least
   re-syncs on the transition.
2. **The actual behaviour question**: what *should* the control show under Venue? I'd argue
   the whole Background group should go inert with a reason ("Venue visualization is active
   — pick a background from the visualization picker"), since none of the three controls
   reach the mounted style. That needs `_pcSync` to consult `_venueSceneOverride` (it's a
   module-scope `let` at 2495, so the `_pc*` block can read it directly) rather than only
   the stored key, plus a `venue` row in `_PC_USES`.

Worth confirming with whoever owns the Venue flow whether Venue is expected to be
user-exitable from the player chrome at all.

## 2. "these settings are global" is true-ish — per-panel overrides shadow them

The refcounting rationale says the settings are global, so N controls would be N ways to
set one value. That's the right conclusion, but the premise has a hole worth knowing about.

`_bgReadSetting(panelKey, key)` (2767) prefers a per-panel override:

```js
panelVal = localStorage.getItem('h3d_bg_' + panelKey + '_' + key);
...
if (panelVal !== null && panelVal !== undefined) return _bgCoerce(key, panelVal);
```

and the renderer reads with a real panel key (`const panelKey = _bgPanelKey(highwayCanvas)`,
8622–8625), while the control reads with `panelKey = null` → looks up the nonexistent
`h3d_bg_null_style` → falls through to the global. So it happens to read the global
correctly, but by falling through rather than by asking for it.

The consequence: if a panel ever has `h3d_bg_panel<N>_style` set, that panel ignores
everything the player control writes, and the control has no way to show that. It's a
**latent** issue — nothing in-app appears to write those keys today, they're only described
in `plugins/highway_3d/CLAUDE.md` — so this isn't a live bug. But two suggestions:

- Pass an explicit sentinel or add a `_bgReadGlobal(key)` helper rather than relying on
  `'h3d_bg_null_' + key` never existing. `null` stringifying into a storage key is the kind
  of thing that works until someone adds a `panelKey` of `null` on purpose.
- Adjust the refcount comment to say the settings are *global with per-panel overrides*, so
  the next person doesn't build on the stronger claim.

## 3. Butterchurn row in `_PC_USES` — right answer, slightly wrong reason

`butterchurn: { intensity: false, reactive: false }` is correct behaviourally, and the
tooltip pointing at Settings → 3D Highway / the Visualizer panel is the right pointer.

The stated reason ("`butterchurn` is not a `BG_STYLES` entry at all — `_bgMountStyle` falls
through to `BG_STYLES.off`") is a bit off from how it works now. `butterchurn` is a real
mounted mode: `_bcSyncMode()` (~8964) runs at the end of `_bgMountStyle` and owns the
controller lifecycle; the *fog-scenery* half falls through to `off`, the visualizer half
doesn't. Practically it means the same thing for this table, but the comment will mislead
whoever next tries to make Intensity map to the Butterchurn canvas opacity — which is a
reasonable future request, and is a real, separate setting the `_bc*` controller owns.

One thing worth checking on hardware if you get the chance: `_bgRebuild()` on a style change
tears down and re-mounts, and `_bcSyncMode` only recreates the controller when
`!bcCtrl || bcCtrl.dead()`. That guard is what keeps the browser audio path from calling
`createMediaElementSource(#audio)` a second time — it's irrevocable per element per
AudioContext, and a second call throws `InvalidStateError` (the caveat is documented at
2204–2345, and the `audioProvider` passed at ~8975 exists precisely to reuse the shared
analyser). Your PR makes style-switching *much* more frequent than it was when it was two
clicks deep in Settings, so a butterchurn → off → butterchurn → off cycle mid-song is now a
realistic user action rather than a theoretical one. I believe the existing guards hold, but
it's the one path in this area where a regression is silent (song audio dies, no console
error on some browsers) and it's newly easy to hit.

## 4. `_pcBindScreenHook` can lose its only chance to bind

`_pcAcquire` (4305) calls `_pcBindScreenHook()` once, then falls into the mount-retry loop:

```js
function _pcAcquire() {
    _pcRefs++;
    _pcBindScreenHook();
    if (_pcMount()) return;
    ...
    const tick = () => {
        ...
        if (_pcMount()) return;
        ...
    };
}
```

`_pcBindScreenHook` (4333) returns silently if `window.feedBack.on` isn't a function yet:

```js
const bus = window.feedBack;
if (!bus || typeof bus.on !== 'function') return;
```

On a cold load where the renderer inits before the bus is wired, that's the only attempt —
the retry tick only calls `_pcMount()`. The control mounts fine, but it's permanently deaf
to chrome rebuilds for the rest of the session, which is exactly the failure the hook was
added to fix (and it's invisible until someone changes screens). One-line fix: call
`_pcBindScreenHook()` at the top of `tick` too, it's already idempotent via the
`if (_pcScreenHook) return` guard.

## 5. `_pcSlot()` doesn't gate on `uiVersion`

`docs/plugin-v3-ui.md` (lines 33–71, and the checklist at 181) makes gating on
`window.feedBack.uiVersion === 'v3'` an explicit rule, and `plugins/tuner/utils/ui.js`
follows it. `_pcSlot()` (4079) only checks that `playerControlSlot` is callable.

In practice this is fine — v2 doesn't expose the function — so it's a consistency nit, not
a bug. But it's a documented checklist item and the tuner is the precedent, so it'll get
flagged eventually. Cheap to align.

## 6. Three copies of the same facts — worth collapsing

`_PC_LABELS` is documented as "kept verbatim in sync with settings.html's `<option>` text",
`_PC_USES` carries a "KEEP IN STEP WITH BG_STYLES" warning, and the test's `EXPECTED_USES`
is a third hand-maintained copy of the capability facts. Each has a comment explaining the
drift risk, which is the honest thing to do — but the drift risk is fixable rather than
manageable.

`BG_STYLES` entries are the natural home for both. Adding `label` and
`uses: { intensity, reactive, why }` to each entry lets `_PC_LABELS` and `_PC_USES` be
derived, and `venue` picks up a row for free (it already has a `BG_STYLES` entry at 3401),
which is half of finding #1. `butterchurn` needs an explicit override either way since it's
mode-not-style.

On the test specifically: the PR body says the expectations are derived from the
`BG_STYLES` bodies "rather than read from `_PC_USES`, so it can't pass by asserting the
table equals itself." True, and a good instinct — but `EXPECTED_USES` is still a literal a
human transcribed once, so it can drift from `BG_STYLES` in exactly the way `_PC_USES` can.
If you want the property the comment is reaching for, scan the source: slice each
`BG_STYLES` entry body the same way `load()` already slices the `_pc*` block, and assert
`intensity === /settings\.intensity/.test(body)` and `reactive === /\bbands\b/.test(body)`.
Then adding a style that reads `bands` without a `_PC_USES` row fails the test instead of
silently shipping an inert control — which is the bug the table's warning comment is asking
a human to prevent.

## 7. Accessibility

The greyed-out affordance is currently `title`-only:

- `title` isn't reliably announced by screen readers and never appears on touch, so on a
  tablet a greyed control is inert with no explanation at all. `aria-describedby` pointing
  at a visually-hidden span would carry the reason properly; the wrapper `title` can stay
  as the mouse affordance.
- The Reactive pill is a toggle button but only sets `aria-disabled` — it needs
  `aria-pressed` reflecting `_on`, otherwise its state is invisible to AT.
- The `<select>` and the range input have `title` but no accessible name
  (`aria-label="Background style"` / `"Background intensity"`).
- Setting `disabled = true` on the slider drops it from the tab order, so keyboard users
  can't reach it to discover *why* it's inert. `aria-disabled` + ignoring input (which the
  handlers already do) keeps it focusable and explainable. The click/change handlers
  already early-return on `disabled`, so most of the work is done.

---

## Small stuff

- `_pcSync` assigns `_pcSel.value` after possibly disabling the `image`/`video` options.
  Programmatically selecting a disabled option works, so if a user's stored style is
  `image` and the image is later cleared, the dropdown shows a selected-but-disabled
  option. Probably fine — worth deciding deliberately rather than by accident.
- Retry gives up after ~3s (12 × 250ms) with no log. With the `screen:changed` hook as
  backstop that's defensible, but a single `console.debug` on the final give-up would make
  "why is there no Background control" answerable from a bug report.
- The test's slice markers mean any code added between `const _PC_LABELS = {` and the
  Factory banner gets evaluated in the sandbox. It's guarded by the marker assertions, but
  the failure mode for the next person adding a helper there is a confusing
  `ReferenceError` rather than "you need to stub this". A one-line note at the top of
  `load()` would save someone ten minutes.

## Suggested priority

1. **#1 Venue desync** — user-visible, and the "dropdown does nothing" symptom will get
   reported as a general bug rather than a Venue-specific one.
2. **#4 bind race** — one line, silent failure.
3. **#7 accessibility** — mostly attributes, and the disabled-control explanation is the
   feature's headline affordance.
4. **#6 single source of truth** — the largest change and the one that retires the whole
   drift class; worth doing before another style lands.
5. **#2 / #3 / #5** — comment and consistency fixes, no rush.

Happy to dig into any of these further, particularly the Butterchurn audio-tap path — that
one has the most history behind it and the least of it is obvious from the code.

Issue #61, opened 2026-06-20
Loop sharing/export


byrongamatos:
> _Restored from `slopsmith/slopsmith#227` — original issue, opened by @byrongamatos on 2026-05-09._
> _[restored-from: slopsmith/slopsmith#227]_

_From `specs/001-slopsmith-platform/tasks.md` — generated retrospectively by spec-kit._

Loop sharing/export — currently scoped to one user's `meta.db`; no documented bundle or QR-code path

---
See `specs/001-slopsmith-platform/spec.md` and `specs/001-slopsmith-platform/analyze.md` for context.
This issue was filed in bulk; close as `wontfix` if no longer relevant.

PR #1017, opened 2026-07-20
fix(sloppak/ws_highway): drums stay in picker and show correct hit count

0-Scr4tch-0:
What

Fixes drums disappearing from the arrangement picker when imported alongside other instruments (e.g. bass), and shows the correct note count instead of "Drums (0)".

The editor strips drum arrangements from the manifest and converts them to drum_tab.json, so a saved drum+bass sloppak has no drum arrangement — only bass plus a drum_tab. The sloppak loader's placeholder-creation trigger was "arrangements list is empty", which only fired for drum-only packs. Broadened to "no drum arrangement exists" so drums get a placeholder even when other instruments are present. The placeholder arrangement also now displays the actual drum_tab hit count instead of 0.

feedpak surface
- This PR does not change how the app reads/writes feedpaks (manifest keys, pack files, folder layout)

Checklist
- CHANGELOG.md [Unreleased] updated (user-visible changes)
- Tests added/updated for new behaviour
- Commits are DCO signed off (git commit -s)

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured drum tab data results in a “Drums” placeholder arrangement when no existing drum/percussion arrangement is present.
  * Fixed displayed note counts for drum-only arrangements to match the underlying drum hit totals.
  * Improved song duration calculation when duration metadata is missing/invalid by deriving it from available drum timing information.

### DIFF
--- a/lib/routers/ws_highway.py
+++ b/lib/routers/ws_highway.py
@@ -478,12 +478,24 @@ async def highway_ws(websocket: WebSocket, filename: str, arrangement: int = -1,
             _evict_audio_cache()
 
         # Send song metadata
+        # For drum-only sloppaks the placeholder arrangement has 0 notes
+        # (drum_tab hits live separately).  Surface the real hit count so
+        # the UI shows e.g. "Drums (1922)" instead of "Drums (0)".
+        _DRUM_KEYWORDS = ("drum", "percussion")
+        _dt_hit_count = 0
+        if is_slop and loaded_slop is not None and loaded_slop.drum_tab is not None:
+            _dt_hit_count = len(loaded_slop.drum_tab.get("hits") or [])
         arr_list = [
             {
                 "index": i,
                 "name": a.name,
                 "smart_name": smart_names[i],
-                "notes": len(a.notes) + sum(len(c.notes) for c in a.chords),
+                "notes": (
+                    _dt_hit_count
+                    if (len(a.notes) == 0 and not a.chords and _dt_hit_count > 0
+                        and any(kw in (a.name or "").lower() for kw in _DRUM_KEYWORDS))
+                    else len(a.notes) + sum(len(c.notes) for c in a.chords)
+                ),
             }
             for i, a in enumerate(song.arrangements)
         ]
--- a/lib/sloppak.py
+++ b/lib/sloppak.py
@@ -895,16 +895,21 @@ def load_song(
                     log.warning("sloppak: drum_tab %r failed validation: %s",
                                 drum_tab_rel, reason)
 
-    # Drum-only sloppak: every GP track was percussion, so it ships a
-    # drum_tab but no pitched arrangements. The highway WS rejects an empty
-    # arrangements list with "No arrangements found" *before* it serves the
-    # drum_tab, leaving the drums unplayable even in the drum highway.
-    # Synthesize a minimal placeholder arrangement so the stream proceeds and
-    # the drum_tab reaches the drum highway. It carries no notes (the guitar
-    # highway just shows an empty board) and, when the manifest omits a
+    # Drum sloppak: when a drum_tab is present but no existing arrangement is
+    # a drum part, synthesize a minimal placeholder so the drums appear in the
+    # arrangement picker and the drum_tab reaches the drum highway.  The editor
+    # strips drum arrangements out of the manifest (converting them to
+    # drum_tab), so without this a drum+bass sloppak would show only Bass in
+    # the picker with no way to reach the drums.  It carries no notes (the
+    # guitar highway just shows an empty board) and, when the manifest omits a
     # duration, derives a song length from the last drum hit so the timeline
     # isn't zero-length.
-    if not song.arrangements and drum_tab_data is not None:
+    _DRUM_KEYWORDS = ("drum", "percussion")
+    _has_drum_arr = any(
+        any(kw in (getattr(a, "name", "") or "").lower() for kw in _DRUM_KEYWORDS)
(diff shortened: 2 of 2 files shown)


gionnibgud:
Evaluated the PR together with the roadmap report. Verdict: **correct fix, right layer** — reader-side compensation in `lib/`, zero feedpak surface touched — and it's future-proof against everything currently planned. Two blockers before merge.

## Future-proofing (checked against #61, the parts-model note, and the editor work)

- **`parts[]`**: `drum_tab` is explicitly preserved as the single-part projection, so this code keeps operating on the same data. Once a real drum part exists in the arrangement list, the placeholder trigger simply stops firing — the code goes dormant for new packs and keeps serving legacy ones. No revert needed.
- **feedpak-spec#61**: binds sound to the song-level `drum_tab` anchor *because* drums have no arrangement entry — it assumes exactly the world this PR fixes. This PR is a prerequisite for it, not a conflict.
- **Editor drums-as-arrangements**: host-side runtime model in the editor plugin, gated on a spec FEP; the editor still strips drums to `drum_tab` on save, so packs reaching this loader don't change shape.

## Blockers

1. **Checklist doesn't match the diff.** "Tests added/updated" and "CHANGELOG updated" are checked, but the PR is two source files — no test, no CHANGELOG entry. `tests/test_sloppak_drumtab_load.py` already has the `_write_dir_sloppak` fixture; two cases cover this PR (~10 lines): a drum+bass pack grows a trailing `Drums` arrangement, and a pack with an existing drum-named arrangement does *not* grow a duplicate. Please also cover the `ws_highway` count-override branch — that's the part that will silently rot untested. And add the `[Unreleased]` CHANGELOG entry.
2. **Drop the keyword sniff from `ws_highway` — the loader already knows which arrangement is the placeholder.** Instead of duplicating `_DRUM_KEYWORDS` + the substring loop in `lib/routers/ws_highway.py`, have `load_song()` record the index it appends (e.g. `drum_placeholder_index: int | None` on `LoadedSloppak` — note `arrangement_ids[i] is None` won't do as the marker, since a manifest entry missing an `id` also lands `None`). The count override then becomes `i == loaded_slop.drum_placeholder_index`: deterministic, no name-guessing, no risk of lighting up an unrelated empty track. That leaves exactly **one** sniff — the loader's "does a drum arrangement already exist?" check — which should live as `is_drum_arrangement_name(name)` in `lib/drums.py`. Name-guessing there is unavoidable today (manifest arrangement entries carry no instrument field), and becomes a one-line key check inside that helper once `parts[]` / feedpak-spec#56 grants one. Keep the keyword list conservative: a false negative costs a cosmetic duplicate placeholder, a false positive re-hides drums — the original bug.

## One reframing request on the report

The report calls the editor's manifest-stripping the "root cause". It isn't a defect — `drum_tab` is the spec's home for drums, so `arrangements[]` *should* exclude them. The loader's synthesized placeholder is the correct projection of song-level drum data into the arrangement list (the same projection `parts[]` will later formalize), not a patch over editor misbehavior. Worth stating it that way so nobody "fixes" the editor later.

## Follow-up filed

#1027 — `extract_meta()` still reads only `manifest["arrangements"]`, so the library index/instrument filters won't show drums for a drum+bass pack even though the picker now does. Out of scope for this PR; the issue includes the fix shape (reuse the same `lib/drums.py` helper — never a third copy of the sniff).

With the tests, the CHANGELOG entry, and the placeholder-index + shared-helper change, this is a merge.

byrongamatos:
## Review — please rebase & rework before this can merge (do not merge as-is)

Thanks for this — the two problems it targets are **real**, but the branch forked **before** #1020 (multi-drum loader) landed on `main`, and once rebased it collides with #1020's model. Details:

**1. Stale base → semantic revert of #1020 (blocker).** This branch's merge-base is `a9be210` (pre-#1020/#1019), so the PR diff *looks* clean but hides that `main` has since rewritten the same two files. Trial-merging onto current `main` auto-merges with no textual conflict but then **fails #1020's grading-invariant test** `tests/test_sloppak_drum_parts.py::test_pointer_entries_never_become_fretted_arrangements` (a `Lead + 2 drum-pointer` manifest expects `song.arrangements == ["Lead"]`; with this change it becomes `["Lead", "Drums"]`). #1020 deliberately keeps `type:"drums"` pointer entries **out** of `song.arrangements` and surfaces them via `song_info.drum_parts`; the name-based `"Drums"` placeholder here re-injects one for exactly the multi-part case #1020 already handles → drums double-surface (once in the main picker collapsing all parts to part0, once via the `drum_parts` sub-picker in #1028).

**2. The single-drum + melodic gap is genuine, but the fix is too broad.** For *one* drum part + a melodic arrangement, #1020 puts a single entry in `drum_parts` while the `drum_part` picker (#1028) only engages for `len > 1`, so that lone drum part isn't selectable. Worth fixing — but scope it to that case via `drum_parts`, **not** by appending a placeholder to `song.arrangements` on a name match. The hit-count root cause (0-note placeholder → "Drums (0)") is correctly diagnosed.

**3. Missing before merge:** no `CHANGELOG` `[Unreleased]` entry, and no loader test.

### To land this
- Rebase onto current `main` (post-#1020/#1019).
- Re-express the single-drum+melodic fix in terms of `drum_parts` (don't add to `song.arrangements`); ideally also populate per-part hit counts in the `drum_parts` payload so the sub-picker shows counts.
- Add a `CHANGELOG` entry + a loader test, and confirm `test_pointer_entries_never_become_fretted_arrangements` stays green.

Happy to re-review once rebased. 🙏

PR #1018, opened 2026-07-20
fix(highway_3d): Venue desync, bind race, and a11y for the player background control

Elit3d:
## What

Follow-up to #1008 (mid-song background controls in the player chrome), addressing post-merge review feedback:

- **Venue desync** — the control read the *stored* background style, not the *effective* one. While the Venue scene override is active, `_bgMountStyle` resolves to `venue` regardless of what the dropdown shows, so picking a style did nothing and looked broken. The whole Background group (dropdown, Reactive, Intensity) now goes inert with a reason on hover while Venue is active, and resyncs correctly when it toggles off.
- **Cold-load bind race** — the `screen:changed` re-mount hook only tried to bind once; on a cold load where the event bus wasn't ready yet, the control could mount but stay permanently deaf to player-chrome rebuilds for the session. It now retries the bind alongside the mount retry.
- **Global-read correctness** — replaced `_bgReadSetting(null, key)` (which worked by relying on a `'h3d_bg_null_<key>'` storage key never existing) with an explicit `_bgReadGlobal(key)` helper, so the shared control's intent to always read the global slot is stated directly rather than implied.
- **Accessibility** — `aria-pressed` on the Reactive toggle, `aria-describedby` pointing disabled controls at a visually-hidden reason (previously title-only, which screen readers announce unreliably and touch doesn't surface at all), and `aria-label` on the style select and intensity slider.
- **v3 UI gate** — `_pcSlot()` now explicitly checks `window.feedBack.uiVersion === 'v3'` per docs/plugin-v3-ui.md, matching the pattern the tuner plugin already follows.

## feedpak surface

## Checklist

- [ ] Commits are DCO signed off (`git commit -s`)

## Summary by CodeRabbit

- **New Features**
  - Background controls in the 3D Highway player now consistently use global background settings.
  - Improved accessibility: disabled-state explanations are exposed to screen readers, and control toggle states are announced correctly.

- **Bug Fixes**
  - Background controls disappear when a non-3D renderer is selected.
  - Background controls are greyed out and inert while a Venue scene override is active, and UI refresh now correctly tracks Venue changes.
  - More robust initialization so the background controls don’t become unresponsive during async load.

- **Documentation**
  - Clarified the unreleased changelog entry for background control behavior under Venue and renderer changes.

- **Tests**
  - Expanded background control coverage, including accessibility and retry behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   style ignores are greyed out with a reason on hover (Custom video and
   Butterchurn use neither; Custom image uses Intensity but not Reactive), so
   a knob is never present-but-inert. The control disappears when a non-3D
-  renderer is selected.
+  renderer is selected. The whole group also greys out while the Venue scene
+  override is active, since none of the three controls reach a mounted style
+  in that mode.
 
 ### Changed
 - **`GET /api/song/{f}?stems=1`** (new, opt-in) — returns the pack's playable stem
--- a/plugins/highway_3d/plugin.json
+++ b/plugins/highway_3d/plugin.json
@@ -1,7 +1,7 @@
 {
     "id": "highway_3d",
     "name": "3D Highway",
-    "version": "3.34.0",
+    "version": "3.34.1",
     "type": "visualization",
     "bundled": true,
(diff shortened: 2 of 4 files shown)


Elit3d:
@coderabbitai review

Issue #340, opened 2026-04-13
[CachyOS/t2-Linux/AppImage] Can't set paths via Settings


mogul:
> _Restored from `slopsmith/slopsmith#556` — original issue, opened by @mogul on 2026-04-13._
> _[restored-from: slopsmith/slopsmith#556]_

When I go to the settings, and click any "Browse" button, nothing happens. Similarly, if I click the box for the path, it's focused, but I'm unable to type in it.

mogul:
_@mogul commented on 2026-04-13:_

I saw you fixed it... Happy to test a new AppImage release if you tag one.

mogul:
_@byrongamatos commented on 2026-04-13:_

Sorry im new to creating releases :) Can you try now?

mogul:
_@mogul commented on 2026-04-14:_

Confirming fixed, thanks! Still no audio or notes showing up when I hit play, but I suspect that's on me.

(For next time: Please make a new tag rather than moving an existing tag to a new hash, to help folks understand what version they've downloaded and are running, and to avoid the appearance of a security issue; see Trivy's recent compromise that targeted people by moving a tag to a malware hash.)

mogul:
_@mogul commented on 2026-04-16:_

(I later confirmed: Yes, the audio issue was on me. Will be collecting a log of my experience and some PRs when I have time to revisit this.)

Issue #335, opened 2026-06-20
GP import: extract chord-diagram fingerings (GP5/GP8) + GP8 chord names


byrongamatos:
GP chord *diagrams* contain per-string fingerings and GP8 carries chord names, but both converters drop them (`fingers=[-1]`; GP8 `name=''`).

## Work
- `lib/gp2rs.py` (GP5) + `lib/gp2rs_gpx.py` (GP8): extract fingerings from GP chord diagrams; GP8: extract chord names.
- Imports then arrive with rich templates the editor now preserves.

Non-blocking; lands after the editor chord-model foundation (got-feedback/feedback-plugin-editor#3) so the data isn't immediately wiped. Part of #334.

PR #1021, opened 2026-07-20
feat(player): drum-part picker for multiple drum charts (feedpak 1.17.0)

ChrisBeWithYou:
**The last mile of the multiple-drum-parts feature** — stacked on #1020 (base `feat/drum-parts-loader`). #1020 taught the loader + highway WS to carry several drum charts (`song_info.drum_parts`, `?drum_part=<id>`, a `part_id` echo on `drum_tab`); this adds the **host-chrome selector** that lets a player *choose* which drummer plays.

## What
A **Drum part** `<select>` sits beside the arrangement switcher (advanced settings), shown only when a song has **2+ drum charts**. Selecting one re-streams that part's tab over the highway WS — exactly like an arrangement switch. Single-drum / no-drum songs hide the row (nothing changes for them).

- **`static/highway.js`** — `reconnect()` gains a `drumPart` arg → `?drum_part=<id>` on the WS URL (mirrors the `arrangement` param); the `song_info` handler populates `#drum-part-select` + shows/hides the row on `length > 1` (parallel to the `#arr-select` block); the `drum_tab` handler carries `msg.part_id` onto `hwState.drumTab` (plugins can read `bundle.drumTab.part_id`) and reflects it as the picker's selected value (stays honest when the server falls an unknown selection back to the primary).
- **`static/app.js`** — `changeArrangement()` gains an optional `drumPart` (forwarded, else the current picker selection is preserved so an arrangement switch keeps the chosen part — parts are song-level); new **`changeDrumPart(id)`** delegates to it (a part switch is the same re-stream, reusing the transition ceremony). Exported on `window`.
- **`static/v3/index.html`** — the `#drum-part-select` row (hidden by default).

**No plugin change** — the drum renderers just draw whatever `drum_tab` streams. Since the WS reconnect is host-owned, the picker lives in host chrome (like the arrangement switcher), so it works regardless of which drum renderer is active.

## Verification
- **Runtime-verified, 10/10** (Playwright, the core player, a real 2-drum pack + a no-drum pack): the picker populates with both parts and shows for the multi-drum song; `song_info.drum_parts` reaches `getSongInfo()`; the primary is pre-selected; selecting the 2nd part **drives `highway.reconnect` with the id and the WS URL carries `?drum_part=drums-2`**; the picker then reflects the server's `part_id` echo; a no-drum song hides the row; no page errors.
- **ESLint 0 errors** (2 pre-existing `max-lines` warnings on these already-large files). JS-only; no pytest touched.

> ⚠️ Stacked on #1020 (`feat/drum-parts-loader`) — merge that first; CI is main-only until this retargets. And the editor + spec sides of the arc (got-feedBack/feedpak-spec#63, got-feedBack/feedBack-plugin-editor #335→#340) complete the authoring end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **Drum-part picker (feedpak 1.17.0 "drums as arrangements").** When a song
+  carries several drum charts, a **Drum part** selector appears beside the
+  arrangement switcher (advanced settings) so a player can choose which drummer
+  to play. Selecting one re-streams that part's tab over the highway WS
+  (`?drum_part=<id>`, mirroring the arrangement switch); the choice persists
+  across an arrangement change, and the picker reflects the server's
+  authoritative part (unknown/absent selection falls back to the primary). The
+  row hides for single-drum and non-drum songs, so nothing changes there. Builds
+  on the loader below; no plugin change needed — the drum renderer just draws
+  whatever tab streams.
 - **Multiple drum parts (feedpak 1.17.0 "drums as arrangements").** The sloppak
   loader now reads `type: drums` arrangement entries carrying per-arrangement
   `drum_tab` file pointers — a song can ship several drum charts (a second
--- a/static/app.js
+++ b/static/app.js
@@ -1150,7 +1150,7 @@ window.feedBack.on('song:ready', () => {
 let _arrBusyGen = 0;
 let _arrBusyTimeout = null;
 
-async function changeArrangement(index) {
+async function changeArrangement(index, drumPart) {
     if (currentFilename) {
         // Tear down any pending fresh-load credits before switching: the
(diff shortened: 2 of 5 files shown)


ChrisBeWithYou:
**Auto-closed, not landed.** This PR was closed automatically when its base branch (`feat/drum-parts-loader` / #1020) was squash-merged to `main` this morning — the picker never actually landed. Verified: `drum-part-select` / `changeDrumPart` / `drum_part` are all absent from `main`, even though #1020's loader is in.

**Re-landed unchanged as #1028** (cherry-picked onto current `main`, runtime-verification carried over). Please review there — this thread is dead.

---

**Reconciliation with #1017** (for review): the two are **complementary, not redundant**. This picker only renders for songs with **2+ drum charts**, so a single drum chart bundled with a melodic instrument (e.g. drums + bass) gets no picker row — and, without #1017's broadened placeholder, no way to reach the drums at all, because #1020 kept the placeholder trigger drum-**only**. So #1017 fills the gap #1020/#1021 leave open. It merges clean on the new `main`, but should pick up a `CHANGELOG` entry + a loader test before it lands.

PR #1022, opened 2026-07-21
fix: Folder Library previews on hover via song_preview (parity with grid/list)

Elit3d:
## What

The Folders view didn't preview songs on hover. This brings it to parity with the
grid and list views.

Rather than add preview code to Folder Library, it wires the folder cards/rows
into the **existing** `song_preview` plugin — the same one that already handles
the grid and list. song_preview looks for `#v3-songs [data-fn]` elements
containing a `[data-v3-play]` descendant; the folder tree renders inside
`#v3-songs`, so once the cards carry that markup its observer picks them up with
**zero changes to song_preview**.

Changes:
- `_songCard` / `_songRow` now emit the standard preview markup — `data-fn` (raw
  filename; was `data-filename`, which song_preview never matched since it keys on
  `dataset.fn`) plus a `[data-v3-play]` surface on the art wrap (cards) / thumb
  (rows).
- Removed the custom hover-preview implementation (dedicated `<audio>`, waveform
  indicator, toolbar toggle, and the backend `has_preview` manifest read) that an
  earlier iteration of this branch had added — song_preview owns preview now.

Credit to @topkoa for pointing out song_preview already exists as a plugin.

## feedpak surface

## Checklist

- [ ] Commits are DCO signed off (`git commit -s`) — **not yet; run `git rebase --signoff main`**


topkoa:
I think we already have this exact functionality in a plugin already https://github.com/got-feedBack/feedBack-plugin-song-preview

Elit3d:
> I think we already have this exact functionality in a plugin already https://github.com/got-feedBack/feedBack-plugin-song-preview

I'd been building a custom hover-preview into Folder Library, but you're right that it already exists as a plugin. Ripping all of that out and instead just giving the folder cards/rows the standard data-fn + data-v3-play markup the grid and list views use. song_preview's observer already watches #v3-songs (and the folder tree renders inside it), so it will preview folder cards on hover with zero changes to song_preview.

Changes are ready for review

PR #1042, opened 2026-07-23
feat(ws): stream the rig library to the client

gionnibgud:
## What

Completes the core reader for source rigs — board [Keys Foundations #4](https://github.com/orgs/got-feedBack/projects/4), "Core reader for source rigs (spec 1.18.0)". Follows #1040, which landed the reading half.

#1040 shipped the **references** and not the thing they reference. `base_rig` and each change's `rig` reached the client, and `LoadedSloppak.rigs` — the library those ids point into — sat server-side with **zero consumers**. A client could learn that a change binds `"grand-piano"` and had no way to find out what that was. Same gap on the drum side: P3 resolved each part's `tones` binding in the loader, and the wire payload was still `{id, name}`.

Two additions:

- **New `rigs` WS message** carrying `rigs.json` whole and verbatim.
- **`song_info.drum_parts[].tones`** — each part's resolved binding (entry `tones`, or the `drum_tones` fallback), omitted when the part binds nothing.

Plus the client-side plumbing: `getRigs()` / `getToneBaseRig()`, `bundle.rigs` / `bundle.toneBaseRig`, and both reset paths clearing them between songs.

## Two decisions worth reviewing

**Not gated on tone changes.** The library ships whenever the pack has one, rather than only when the active arrangement has `tone_changes`. A pack can bind sound to its drum parts alone via `drum_tones` and still need the library — gating on arrangement tones would strand exactly that case. Covered by `test_rigs_stream_even_without_arrangement_tones`.

**Verbatim, whole.** No per-rig filtering or resolution. Core selects no realization and applies no `intent.gm` floor, so the consumer that voices the part needs the entire block to make that choice — including the unknown `role`/`engine`/`kind` values and `ext` namespaces §7.9 obliges a Reader to preserve. Sending only "the rigs this arrangement references" would also break the drum case above.

## Compatibility

Packs that bind no rig are **byte-identical on the wire**: no `rigs` message, no `base_rig`, no `rig`, no `tones` key on a drum part. `test_unbound_pack_keeps_the_legacy_payload_shape` asserts exactly that.

## Verification

- Python **2802 passed / 4 skipped** (+6 new in `tests/test_highway_ws_rigs.py`, end-to-end through the real WS via `TestClient`).
- JS **1343 passed / 0 failed**.
- `tools/check_spec_conformance.py`: **OK**.

The new tests cover the pairing that is the actual point — that the ids in `tone_changes` and in a drum part's `tones` resolve against ids present in the `rigs` frame.

## feedpak surface

- [ ] This PR does **not** change how the app reads/writes feedpaks
- [x] …or it does, and the spec change landed first — **spec 1.18.0 / [feedpak-spec#61](https://github.com/got-feedBack/feedpak-spec/issues/61), already merged.** This PR adds no new manifest key; it forwards what #1040 already reads. Gate passes against spec HEAD.

## Checklist

- [x] `CHANGELOG.md` `[Unreleased]` updated

---

Note for whoever merges: `CHANGELOG.md` is CRLF in this repo — worth knowing before editing it with a tool that normalises line endings, which silently rewrites all ~660 lines.

Also open and independent: #1041 (routes `_load_rigs_file` through the shared path helper).


gionnibgud:
Superseded by #1044 — same single commit, reopened from a fresh branch to drop this PR's force-push history (an early push briefly included a local docker-compose.override.yml; it was amended out, but the old SHAs stay linked from this PR's timeline). No code difference.

PR #1045, opened 2026-07-25
feat: expose enriched song metadata to plugins

JMcG1:
## What

Adds a generic Plugin Metadata API: any plugin can now get album, year,
genre, and MusicBrainz identifiers (recording/release/artist MBID, ISRC)
for the currently-playing or any known song — data the server already
computes via its MusicBrainz/AcoustID enrichment pipeline, but that
never reached a plugin before.

Two delivery surfaces, one shared resolver:
- The highway WebSocket's `song_info` frame gains one new optional key,
  `metadata`.
- A new `GET /api/song/{filename}/metadata` REST route serves the same
  shape for a song that isn't currently playing.

Both call the same `lib/plugin_metadata.py::plugin_metadata_for()` —
there is exactly one place this assembly happens, so the two surfaces
can't drift apart in shape or in the rules that gate MusicBrainz data.

## Why this is generic, not Song-Background-Manager-specific

This started as an investigation into what SBM (a community plugin)
would need for album art/MusicBrainz lookups. The server already had
everything — playback metadata, an enrichment pipeline with confirmed
MusicBrainz matches — but exposed none of it to any plugin. Rather than
build a one-off channel for one plugin, this exposes the same
information generically: any current or future plugin (visualizers,
scrobblers, tagging tools, alternate art sources) gets it through the
same public surface. SBM itself is not modified by this PR.

## Backwards compatibility

Purely additive:
- No existing field, on any existing route or WS frame, is removed,
  renamed, or reinterpreted.
- The only new surface is one optional `metadata` key on `song_info`
  and one new REST route.
- The `metadata` object has a stable set of documented keys and uses
  consistent empty or null defaults, allowing consumers to read fields
  safely without checking whether each individual key exists.
- An old plugin that never reads `metadata` behaves identically before
  and after this change. A new plugin against an old server just sees
  `metadata` as `undefined`/absent.

## How it works

- `plugin_metadata_for(filename, *, base_metadata=None)` reads the
  existing `songs` cache (`MetadataDB.pack_fields`) and the existing
  `song_enrichment` table (`MetadataDB.get_enrichment`) — the existing
  indexed MetadataDB lookups, no new tables, no new query shapes, and
  no network access anywhere in this path.
- `base_metadata` lets a caller that already has a loaded `Song` (the
  WebSocket handler, mid-playback) pass its `album`/`year` straight
  from the sloppak manifest instead of waiting on the next library
  scan to populate the cache — resolved per field
  (`base_metadata → songs cache → safe default`), never all-or-nothing.
  The REST route has no loaded `Song` to draw on, so it stays
  cache-only, exactly as before this parameter existed. (Note: this
  means `album`/`year` can briefly differ between the two surfaces for
  a song that hasn't been scanned yet — documented in
  docs/PLUGIN_METADATA_API.md.)
- MusicBrainz identifiers/ISRC are surfaced **only** when the song's
  enrichment match is `matched` or `manual` — a `review`/`failed`/
  `unscanned` row's candidate data is a proposal, not a confirmed
  identity, and is never leaked as if it were one. This mirrors the
  same gate `routers/art.py`'s cover-art candidate assembly already
  applies.

## Why no duplication

Before this PR, adding metadata to both the WS payload and a REST route
independently would mean the two could silently disagree (e.g. one
surface's gating logic for MusicBrainz identifiers drifting from the
other's over time). Routing both through one function makes that
structurally impossible rather than a matter of remembering to keep
two implementations in sync.

## Testing

- `tests/test_plugin_metadata.py` — pure unit tests against a real
  `MetadataDB` (no FastAPI, no network): unknown-song defaults, year
  coercion edge cases, enrichment gating (matched/manual/review/failed/
  unscanned), `base_metadata` resolution (overrides cache, blank/zero
  falls through, partial base_metadata still pulls missing fields from
  cache, enrichment unaffected either way), JSON round-trip.
- `tests/test_plugin_metadata_api.py` — WS/REST integration tests:
  route precedence (the new `/metadata` route can't be shadowed by the
  existing bare `GET /api/song/{filename}` catch-all), missing-song
  handling, and confirmation that every pre-existing `song_info` field
  is unchanged by this addition.
- Full existing WebSocket regression suite (`test_highway_ws_authors`,
  `test_highway_ws_instrument_routing`, `test_highway_ws_notation`,
  `test_ws_highway_disconnect`) passing with no regressions.

See `docs/PLUGIN_METADATA_API.md` for the full field reference and
`docs/PLUGIN_METADATA_FINAL_VERIFICATION.md` for the pre-merge
verification pass.

## feedpak surface

- [x] This PR does **not** change how the app reads/writes feedpaks
      (manifest keys, pack files, folder layout) — it only reads
      fields the existing loader/scanner already parse.

## Checklist

- [x] `CHANGELOG.md` `[Unreleased]` updated

## Summary by CodeRabbit

* **New Features**
  * Added a Plugin Metadata API to expose per-song album, year, genre, ISRC, and MusicBrainz identifiers.
  * Metadata is available via a new REST endpoint and optionally included in highway WebSocket `song_info`, using a stable, additive response shape with safe null/empty values when unavailable.
* **Documentation**
  * Published API documentation covering payload shape, versioning/compatibility, enrichment gating, error behavior, and performance notes.
  * Added final verification documentation for the feature.
  * Updated the changelog entry.
* **Tests**
  * Added unit and integration tests covering REST/WebSocket delivery, enrichment match states, missing-song behavior, and backward compatibility.


JMcG1:
Ready for review when you get a chance. I’m holding off on merging until approval comes through.

PR #1066, opened 2026-08-23
fix(security): add worker-src to CSP so alphaTab-based plugins can render

carochacs:
## Summary
- Adds a `worker-src 'self' blob:;` directive to the site-wide Content-Security-Policy in `server.py`.

## Root cause
The baseline CSP headers added in 81837ec (#44, stored-XSS fix) covered `script-src`, `style-src`, `img-src`, `font-src`, `media-src`, and `connect-src`, but never set `worker-src`. Per CSP fallback rules, workers without an explicit `worker-src` fall back to `script-src`, which allows `'self'`/`'unsafe-inline'`/`https:` but **not** `blob:`.

alphaTab (used by the `tabview` and `staffview` plugins) spawns its rendering worker from a `blob:` URL. Without `worker-src`, browsers silently blocked that worker, so both plugins rendered a blank canvas — for every song and every arrangement, not anything instrument-specific.

Reported as "can't see the Sax arrangement in staffview or tabview" for the song *Money*. Reproduced the identical blank render on the *Lead* arrangement of the same song, confirming this is a general regression from the CSP change introduced in #44, not anything about the Sax arrangement's data.

## Fix
Add `worker-src 'self' blob:;`, matching the `blob:` allowance already granted to `img-src`/`media-src` for the same reason (plugin-loaded CDN assets).

## Test plan
- [x] Rebuilt the Docker image and restarted the container with the fix applied.
- [x] Loaded the song *Money*, selected the Sax arrangement, switched the active renderer to Tab View — full tablature now renders (measure numbers, tempo markings, tab lines) instead of a blank canvas.
- [x] Verified Staff View renders as well.
- [x] Confirmed the failure was general (not Sax-specific) by reproducing the same blank render on the Lead arrangement before the fix, and the same successful render after.

## Summary by CodeRabbit

- **Security**
  - Added protection against unsafe XML imports, malicious archives, stored HTML injection, and clickjacking.
  - Added security headers to all responses.
  - Demo mode now blocks lyric-saving actions.

- **Bug Fixes**
  - Improved errors when song packages fail to load without exposing internal details.
  - Preserved chart events occurring before time zero.
  - Improved chart anchor defaults before the first anchor.

- **Enhancements**
  - Guitar imports now preserve pick direction and capo settings.
  - Library card actions support song-specific labels and icons.
  - Updated version to 0.3.0-alpha.2.

- **Documentation**
  - Added project setup, plugin, contribution, and fork-maintenance guidance.


carochacs:
Closing — opened against the wrong repo. This branch was cut from `get-flashbacks/feedBack`'s `main` (which is ahead of this upstream repo), so the diff against `got-feedBack/feedBack:main` pulled in unrelated upstream-vs-fork history instead of just the one-line CSP fix. The bug this fixes only exists in `get-flashbacks/feedBack` right now. Reopening the correct PR there.

PR #1068, opened 2026-08-31
fix: preserve Guitar Pro pick direction

### LINKED ISSUE #1054: GP import discards strumming direction (pick_direction stays unset for every imported note)
## Summary
`lib/song.py` already has a fully-wired field for this — `Note.pick_direction` (default `-1`), serialized as `pkd` on the wire (`song.py:267-268`) and `pickDirection` in RS-XML (`song.py:1110`) — but neither Guitar Pro importer ever populates it:

- `lib/gp2rs.py` (GP3/4/5 binary import via PyGuitarPro) — no reference to stroke/pick direction anywhere in the file.
- `lib/gp2rs_gpx.py` (GP6/7/8 GPX/XML import) — same, no reference.

Every note imported from a `.gp3/.gp4/.gp5/.gpx` file ends up with `pick_direction = -1` (unset), even when the source file authored an explicit up/down stroke.

Separately, `gp2rs.py:1224` writes a `strum="down"` attribute on chord elements, but it's hardcoded unconditionally for every chord regardless of the source data — it doesn't read anything from the file either. That's a different, chord-level legacy attribute, not a substitute for real per-beat direction.

## The source data is actually there to read
Checked the `guitarpro` (PyGuitarPro) library `gp2rs.py` already depends on: `Beat.effect` (a `BeatEffect`) exposes both `stroke: BeatStroke` and `pickStroke: BeatStrokeDirection` (an up/down/none enum) directly — confirmed via `inspect.signature(guitarpro.models.BeatEffect.__init__)`. So for GP3-5 this isn't a "library doesn't expose it" situation; the value is sitting on the parsed beat object and is simply never read. GPX (GP6+) has the equivalent `<Stroke><Direction>` element in its own beat/XML schema.

## Ask
Wire `beat.effect.pickStroke` (GP3-5) / the GPX `<Stroke><Direction>` element (GP6+) into `Note.pick_direction` in both `gp2rs.py` and `gp2rs_gpx.py`, so strumming direction survives import instead of silently dropping to unset. `strum_group` (the rake/strum-gesture grouping key next to `pick_direction` in `song.py`) may also be relevant if a beat's stroke applies across the whole chord rather than per-note — worth checking during implementation whether both fields need populating or just `pick_direction`.

## Context
Found while investigating a related question about why feedpakr flags "two notes on the same string" during GP import (unrelated root cause, already understood — that one's a wire-format encoding artifact for keys/drums/multi-voice guitar, not a bug). Verified not a duplicate of any open issue (searched "strum", "pick direction", "pickDirection", "stroke").

raphaelmorrissy:
## What

Preserve explicitly authored Guitar Pro pick/strum direction during import instead of leaving every converted note's `pickDirection` unset.

- maps PyGuitarPro `pickStroke` and legacy `stroke.direction` values to feedBack's `down = 0` / `up = 1` convention
- reads GP6/7/8 GPIF `<Stroke><Direction>` values
- writes `pickDirection` for standalone notes and every note in a chord
- leaves missing or unknown directions unset (`-1`) and does not fabricate `strum_group` values
- documents the behavior in the changelog and regression tests

Closes #1054

## feedpak surface

- [ ] …or it does, and the spec change landed first via the [FEP process](https://github.com/got-feedback/feedpak-spec/blob/main/CONTRIBUTING.md) — FEP / spec PR: `got-feedback/feedpak-spec#___`

## Checklist

## Testing

- `391 passed` across the GP converters, XML encoding, and song model suites
- Full Python suite: `2769 passed, 32 skipped`; 12 unrelated Windows/environment failures (symlink privileges, path/EOL, and packaging checks)
- `py_compile` and `git diff --check` pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Guitar Pro imports now preserve authored up/down picking directions across GP3–GP8 files.
  - Picking directions are retained for both individual notes and chord notes.
  - E Standard retuning now keeps tuning metadata consistent across all arrangement files.
  - Retuning fails clearly if metadata cannot be updated everywhere, preventing inconsistent results.

- **Tests**
  - Added coverage for picking-direction imports, legacy direction data, chord handling, and GP8 files.


raphaelmorrissy:
@coderabbitai review

raphaelmorrissy:
@coderabbitai review

raphaelmorrissy:
@ChrisBeWithYou, when you have a moment, could you please review this GP importer fix? CodeRabbit reports no actionable comments, all 5 pre-merge checks pass, and the related converter tests are green. Thanks!

